PR #65, opened 2013-02-22
Fixed to handle git submodules

unluckypixie:
Hi,

I have a project which heavily uses git submodules (around 80) and I wanted the branch name from each submodule to use in the name of the package for the submodule.  I got this working many months back with an old version of your plugin but the recent changes to git have changed the way .git directories work for git submodules (they are now a file with a path to where the actual directory is), I have been putting off upgrading the version of git my team was using but it has now become an issue so I have updated your plugin to support the new git submodule format.

That was the primary purpose of my commit but I have also done a couple of other things (hope you don't mind!) ...

1) Slightly changed the logging so that there is a "alwaysLog" and "log" function.  The always log function logs messages even when verbose is false and the "log" only when verbose is true.  I have made the path to the git directory and the loaded properties always log now (thought it's easy to change if you disagree).
2) The new feature to inject the properties into reactor projects really slows down my build (I have 80 is projects and have to run the plugin on each one).   So I have added a "injectAllReactorProjects" parameter (which defaults to the current behaviour of "true").  If you set that to false it only injects into the current project like it used to.
3) As part of the git submodule fix I discovered that the searching of the heirarchy of projects for the .git directory simply does not work (I have a couple of poms in a heirarchy of directories that share one .git project and it didn't work with them).  This seems to be down to project.getProject() always returning NULL.  I have added some code so that if getParent() returns NULL but getParentArtifact() does not then I search the reactor projects for the one that matches the parentArtifact and use that.  I think the fact that getParent() returns NULL is a bug in Maven so this is kinda a workaround and if it did return a project that one would be used.

All of the changes (apart from the log verbosity change) should give the same current behaviour with their default settings.  I have tested the plugin with git 1.7 and git 1.9 (between those two the git directory format was changed) and it works fine for me with both.  If you accept this plugin I would appreciate if you could get it up into the maven repos (as 2.15?) asap so I can update my projects to use this version!

### DIFF
--- a/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
+++ b/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
@@ -92,8 +92,17 @@ public class GitCommitIdMojo extends AbstractMojo {
    * @parameter expression="${reactorProjects}"
    * @readonly
    */
-  private List reactorProjects;
+  private List<MavenProject> reactorProjects;
 
+  /**
+   * Tell git-commit-id to inject the git properties into all
+   * reactor projects not just the current one.
+   *
+   * @parameter default-value="true"
+   */
+  @SuppressWarnings("UnusedDeclaration")
+  private boolean injectAllReactorProjects;
+  
   /**
    * Specifies whether the goal runs in verbose mode.
    * To be more specific, this means more info being printed out while scanning for paths and also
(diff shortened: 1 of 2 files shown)


kwin:
This should definitely not be the default (i.e. it should be disabled by default, because that will propagate all(!) properties to all maven modules). See https://github.com/ktoso/maven-git-commit-id-plugin/issues/108.

sfisque:
there appears to be an edge case bug.  i have a project that is originally PHP, and in a sub directory under the root of the original sandbox, we have a new spring-boot conversion, that is pom driven.  the original project (being php) has no pom and is not "built" but merely deployed "as is".  i found that i had to set **injectAllReactorProjects** to **true**, otherwise the properties were not expanded during resource filtering (vars remained ${git.var.name} rather than the expanded value.  this seems counter intuitive since the project only has a single pom, but it is located in a subdirectory below the git repository root.

PR #135, opened 2014-10-30
Improve gitdir handling with submodules

maiergre:
Thanks for the plugin! 

This pull req fixes two issues I encountered:
- When using git submodules some older git version (e.g., 1.7.9, used by Ubuntu 12.04 LTS) use absolute path in the .git gitdir reference. I've added support for absolute path.
- When searching for the gitdir the plugin should search the parent _directory_ not the parent project. Esp. with submodules searching the parent project is not correct since the parent might be in a different git repo.

### DIFF
--- a/src/main/java/pl/project13/maven/git/GitDirLocator.java
+++ b/src/main/java/pl/project13/maven/git/GitDirLocator.java
@@ -17,8 +17,6 @@
 
 package pl.project13.maven.git;
 
-import com.google.common.base.Optional;
-import org.apache.maven.artifact.Artifact;
 import org.apache.maven.project.MavenProject;
 import org.eclipse.jgit.lib.Constants;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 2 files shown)


klion26:
Hi thanks for the great project, so that we can track the commit message in the distribute env easily.

I have a question about searching for .git in the parent directory: do we need to add an option to not search the parent directory when finding the `.git` dir?   I am not sure if this is a normal case or not, I encountered this in some testing processes. thanks.

Now we will search all parent directories from the current directory to find the `.git` directory, but when I configure `failOnNoGitDirectory` to `true` and there is a .git directory in a parent directory, the compilation will not report an error and fail, but the `.git` directory founded in the parent directory will be used as the information of the project.

The background of this case is that I am contributing to an open-source project. this project wants to report an error without `.git` under the usual process, but the project also provides a zip package download method to compile without `.git`. When I tested it, I moved the `.git` directory to the parent directory of the project and found it passed the compilation and did not report any error.

Issue #149, opened 2014-12-15
git describe default -dirty is not set by default


Martin-Wegner:
And again... The git describe default for a dirty version (-dirty) is not set by default.

TheSnoozer:
The problem should be resolved and was included in 2.1.13

PR #150, opened 2014-12-18
Fix for https://github.com/ktoso/maven-git-commit-id-plugin/issues/149

TheSnoozer:
As discussed in https://github.com/ktoso/maven-git-commit-id-plugin/issues/149
Using "...-dirty" as default dirty-marker.


ktoso:
Merged this in manually.

PR #152, opened 2015-01-06
Feature/issue 51 update properties file when needed only

apogrebnyak:
Checks if properties file needs updating


ktoso:
Thanks for the contrib!

Issue #108, opened 2014-05-05
injectAllReactorProjects overwrites non-git properties in reactor projects


tnaroska:
The option injectAllReactorProjects should set the git-commit-id related properties in all reactor projects. Right now the code does not only propagate the git properties, but *all* properties set in the current project (i.e. in the pom.xml's <properties> element).
This can overwrite properties completely unrelated to git in other projects leading to unexpected behavior.

kwin:
This is a problem e.g. with the maven-scr-plugin and the maven-bundle-plugin, because they communicate via properties. If those properties are propagated across module boundaries this leads to all kind of errors.

PR #157, opened 2015-03-09
only propagate git properties to other maven modules

kwin:
fixes https://github.com/ktoso/maven-git-commit-id-plugin/issues/108 by only overwriting properties which are set by the maven-git-commit-id-plugin. Also disables the propagation of properties by default.


ktoso:
Thanks, great contrib! I agree with your findings and the fix, thanks again.

PR #161, opened 2015-03-16
Update Guava to 18.0

mkordas:
- Remove usage of deprecated Closeables.closeQuietly() method

Deprecation reason:
>This method has few valid use cases and encourages misuse by making it easy to do the wrong thing. Among other things, it may swallow exceptions that really should be thrown, such as exceptions thrown when closing an output stream: this often involves flushing buffered data to the final output destination and as such it is just as important to throw the exception thrown when closing as it is to throw an exception thrown by a call to a write method. This method is scheduled to be removed in Guava 16.0.

All usages were replaced by `Closeables.close()`

From Javadoc:
>```java
public static void close(@Nullable
         Closeable closeable,
         boolean swallowIOException)
                  throws IOException
```
>Closes a Closeable, with control over whether an IOException may be thrown. This is primarily useful in a finally block, where a thrown exception needs to be logged but not propagated (otherwise the original exception will be lost).
If swallowIOException is true then we never throw IOException but merely log it.

>Example:
```java
   public void useStreamNicely() throws IOException {
     SomeStream stream = new SomeStream("foo");
     boolean threw = true;
     try {
       // ... code which does something with the stream ...
       threw = false;
     } finally {
       // If an exception occurs, rethrow it only if threw==false:
       Closeables.close(stream, threw);
     }
   }
```

- Update com.google.guava:guava to 18.0


mkordas:
We have the problem with `Closeables.closeQuietly(Reader)` used at `src/main/java/pl/project13/maven/git/GitCommitIdMojo.java:[584,18]` that was deprecated in Guava 14.0 and removed in Guava 16.0. I'll deal with that. 

>This method has few valid use cases and encourages misuse by making it easy to do the wrong thing. Among other things, it may swallow exceptions that really should be thrown, such as exceptions thrown when closing an output stream: this often involves flushing 	buffered data to the final output destination and as such it is just as important to throw the exception thrown when closing as it is to throw an exception thrown by a call to a write method. This method is scheduled to be removed in Guava 16.0.

ktoso:
Sure, thanks :)

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
```diff
@@ -631,21 +633,26 @@ static Properties readJsonProperties(@NotNull File jsonFile) throws CannotReadFi
...
-        final InputStreamReader reader = new InputStreamReader(fis, Charsets.UTF_8);
-        closeable = reader;
-
-        final ObjectMapper mapper = new ObjectMapper();
-        final TypeReference<HashMap<String,Object>> mapTypeRef =
-          new TypeReference<HashMap<String,Object>>() {};
-
-        propertiesMap = mapper.readValue(reader, mapTypeRef);
+        boolean threw = true;
+        try {
```
ktoso: It could just be one `try` block. Would you making it so?

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
```diff
@@ -663,24 +670,27 @@ static Properties readProperties(@NotNull File propertiesFile) throws CannotRead
     Closeable closeable = null;
 
     try {
-      final FileInputStream fis = new FileInputStream(propertiesFile);
-      closeable = fis;
+      boolean threw = true;
+      try {
```
ktoso: This also can be just one try block.

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
mkordas: It cannot, as [Closeable.close()](http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/io/Closeables.html#close(java.io.Closeable,%20boolean)) declares `throws IOException` and this exception needs to be repacked into unchecked `CannotReadFileException`.

ktoso: Oh I see. Got too used to closeQuietly I guess.
Thanks

mkordas: The clean solution would be to use try-with-resources, but for that we need Java 7.

PR #175, opened 2015-03-30
Update maven-plugin-api to 3.3.1

mkordas:
The latest version of `org.apache.maven:maven-project` is still 2.2.1, so the properties had to be split.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
   </scm>
 
   <properties>
-    <maven-plugin-api.version>2.2.1</maven-plugin-api.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -69,7 +69,7 @@
   </scm>
 
   <properties>
-    <maven-plugin-api.version>2.2.1</maven-plugin-api.version>
+    <maven-plugin-api.version>3.3.1</maven-plugin-api.version>
```
ktoso: is this still used?

PR #178, opened 2015-04-05
Add git.build.host property

sergey-podolsky:


ktoso:
Hi @sergey-podolsky, it's an interesting idea. 

I'm not sure this should be enabled by default though - sometimes you may not want to expose to the shipped jar the IP of where it was built - I mean, why tell people where your CI is if it's publicly accessible for example. For internal use then sure - but let's hide it behind an option, sounds ok?

sergey-podolsky:
Hi @ktoso,
I have simplified the things. Now it returns just host name, without IP address.
The primary intention of this feature is to provide an ability to track the build server, and to distinguish whether it was a CI build, or developer's build. Currently there is no way to check whether the build is a genuine CI build rather than someones sandboxie build. Existing properties do not provide you any clues, especially if git user uses anonymous checkout mode.
Therefore, IP address is not really essential here, especially if it is used inside of intranet. Host name should be enough.

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
```diff
@@ -472,6 +476,16 @@ void loadBuildTimeData(@NotNull Properties properties) {
     put(properties, BUILD_TIME, smf.format(buildDate));
   }
 
+  void loadBuildHostData(@NotNull Properties properties) {
+    String buildHost = null;
+    try {
+      buildHost = InetAddress.getLocalHost().getHostName();
+    } catch (UnknownHostException e) {
+      log("Unable to get build host");
```
ktoso: Maybe more informative like so: 

```
"Unable to get build host, skipping " + BUILD_HOST + " value..."
```

ktoso: And indeed don't set it if it's `null`.

sergey-podolsky: I have updated the error message to be more verbose.

> And indeed don't set it if it's null.

We have to put null value explicitly, because null is processed into "Unknown" later. I think it would be better to have "Unknown" value in the resulting property rather than omitting the property completely from the list of properties. Someone may get confused in case if property does not exist for some non-obvious reason.

ktoso: Oh OK. I agree with having unknown as the value. Sounds good.

I still do not want to expose any IP addresses unless explicitly turned on
though. Some of the companies using this plugin are very sensitive about
this kind of stuff.

sergey-podolsky: IP address is not exposed anymore, only host name.

src/test/java/pl/project13/maven/git/GitCommitIdMojoTest.java
```diff
@@ -27,6 +27,7 @@
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Paths;
 import java.util.Map;
```
TheSnoozer: Unfortunately the java.nio.file got introduced with JDK 7 and causes the travis build to fail for Java 6.

[ERROR] /home/travis/build/ktoso/maven-git-commit-id-plugin/src/test/java/pl/project13/maven/git/GitCommitIdMojoTest.java:[30,20] package java.nio.file does not exist
[ERROR] /home/travis/build/ktoso/maven-git-commit-id-plugin/src/test/java/pl/project13/maven/git/GitCommitIdMojoTest.java:[224,43] cannot find symbol
symbol  : variable Paths
location: class pl.project13.maven.git.GitCommitIdMojoTest
[ERROR] /home/travis/build/ktoso/maven-git-commit-id-plugin/src/test/java/pl/project13/maven/git/GitCommitIdMojoTest.java:[235,43] cannot find symbol
symbol  : variable Paths
location: class pl.project13.maven.git.GitCommitIdMojoTest

ktoso: Yeah... I'm thinking that maybe it's time to drop JDK6... I mean, JDK7 is OOL in a second.

Let's drop 6 I think.

TheSnoozer: I think we are save moving to Java 7. Even for JDK7 the public EOL is reached after April 2015:
http://www.oracle.com/technetwork/java/javase/downloads/eol-135779.html

Only moving to JDK8 would be a giant leap....
However we should include this info in the readme and I would suggest to include this in a major release (2.2.0) instead of a minor version (2.1.14).

ktoso: We don't really need JDK8 features and some companies are uber-slow in upgrading :-)
I agree that we should bump such dependency during a `2.2.0` :)

sergey-podolsky:
Reverted changes that required Java 7 features.

ktoso:
Awesome, thanks :-)

Issue #96, opened 2014-02-26
Shallow depth returns error


fhanik:
git clone --depth=50 --branch=develop git://github.com/cloudfoundry/login-server.git cloudfoundry/login-server
    git checkout -qf d6c2aee39d95c1ff922b6f1a427424527876e3aa
    mvn -U install --quiet -DskipTests=true

yields the following error 

    [ERROR] Failed to execute goal pl.project13.maven:git-commit-id-plugin:2.1.9:revision (default) on project cloudfoundry-login-server-common: Execution default of goal pl.project13.maven:git-commit-id-plugin:2.1.9:revision failed: Unable to find commits until some tag: Walk failure. Missing commit edebec08c1ecac85968010a0d93f71dd33b44bb8 -> [Help 1]

If I increase the depth to 500, the build works fine. is there a way to not fail the build for a shallow depth?

example
https://travis-ci.org/cloudfoundry/login-server/builds/19687743

TheSnoozer:
Hopefully was fixed with https://github.com/ktoso/maven-git-commit-id-plugin/pull/179.
At least I can not repduce the issue (I had the same back in the days, so I know how to reproduce).

Thanks for reporting & thanks for the support to actually track the problem down.

PR #179, opened 2015-04-07
JGit and other updates

katox:
Hi,
thanks for the plugin. It usually does the job well but it seems to fail with shallow git clones (used by CI servers). See issue #96. 

Updating plugin's jgit version seems to alleviate the problem.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
   </scm>
 
   <properties>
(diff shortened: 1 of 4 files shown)


ktoso:
Always happy to update jgit, let's do it.
Thanks a lot for the contrib @katox

Issue #181, opened 2015-04-08
Maybe introduce a demo-profile inside the pom


TheSnoozer:
I just was thinking if maybe introducing a demo-profile inside the pom is somehow a good idea.
Currently there is an example of the plugin config included, however it's a huge block of comments and currently is barely usable.

Maybe just migrate that to a profile:
```
<profile>
      <id>demo</id>
      <build>
        <plugins>
          <plugin>
            <groupId>pl.project13.maven</groupId>
            <artifactId>git-commit-id-plugin</artifactId>
            <!-- <version>2.1.13</version> -->
            <version>${project.version}</version>
            <executions>
              <execution>
                <goals>
                  <goal>revision</goal>
                </goals>
              </execution>
            </executions>
            <configuration>
              <verbose>true</verbose>
              <skip>false</skip>
              <prefix>git</prefix>
              <dotGitDirectory>${project.basedir}/.git</dotGitDirectory>
              <generateGitPropertiesFile>true</generateGitPropertiesFile>
              <generateGitPropertiesFilename>target/testing.properties</generateGitPropertiesFilename>
              <dateFormat>dd.MM.yyyy '@' HH:mm:ss z</dateFormat>
              <useNativeGit>false</useNativeGit>
              <abbrevLength>7</abbrevLength>
              <format>properties</format>
              <gitDescribe>
                <skip>false</skip>
                <always>false</always>
                <abbrev>7</abbrev>
                <match>*</match>
                <dirty>-DEVEL</dirty>
                <forceLongFormat>false</forceLongFormat>
              </gitDescribe>
              <excludeProperties>
                <excludeProperty>git.commit.*</excludeProperty>
                <excludeProperty>git.remote.origin.url</excludeProperty>
              </excludeProperties>
              <failOnNoGitDirectory>false</failOnNoGitDirectory>
              <failOnUnableToExtractRepoInfo>false</failOnUnableToExtractRepoInfo>
            </configuration>
          </plugin>
        </plugins>
      </build>
    </profile>
```

Demo-run would be possible with:
```
mvn clean install -Dmaven.test.skip=true && mvn clean package -Pdemo -Dmaven.test.skip=true
```

Also add a comment that the plugin could also be included at it's current position.

TheSnoozer:
fixed with #184.

PR #184, opened 2015-04-10
Introduce a demo-profile inside the pom, see #181

TheSnoozer:
As suggested in https://github.com/ktoso/maven-git-commit-id-plugin/issues/181, I would like to introduce a demo-Profile where anybody can just run the plugin right away.

This would speed up to live test wheather something is broken under some conditions or not. I used this in one of my own plugins to discover some nasty bugs.

However, this will not avoid writing Testcases. This should only help to track down a problem and then writing test cases for it.
For my own plugin I once had included it in the regular-build (so run everytime for everytime), However when i screwed up, I had no chance to re run ``mvn clean install`` without removing the config of the plugin. Therefore I needed to moved it to a different profile where I could choose if I would like to run it or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,11 @@ But I highly recommend using only stable versions, from maven central... :-)
 </pluginRepositories>
 ```
 
+If you just would like to see what the plugin can do, you can clone the repository and run
+```
+mvn clean install -Dmaven.test.skip=true && mvn clean package -Pdemo -Dmaven.test.skip=true
+```
+
 Using the plugin
 ----------------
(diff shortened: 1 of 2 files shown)


TheSnoozer:
Hi @ktoso,
I wasn't sure if I just can pull my own change in.
I'll just go ahead and if you don't like just let me know :)

ktoso:
Great idea and good to merge commits whenever you feel a PR is good to go :-)

ktoso:
Let's keep doing things as PRs so we have a chance to discuss of something more up-for-debate :-)

PR #192, opened 2015-05-05
Added includeProperties and additionalProperties as mojo parameters.

Cousnouf:
Hello Konrad,

I'm a user of your plugin and I thought it might be useful to add the two parameters I did. They allow a little bit more flexibility and then to put everything in a same file for instance.

In my case I needed only this configuration:

        <configuration>
          <dotGitDirectory>${project.basedir}/../.git</dotGitDirectory>
          <generateGitPropertiesFile>true</generateGitPropertiesFile>
          <generateGitPropertiesFilename>src/main/resources/myAppInfo.properties</generateGitPropertiesFilename>
          <includeProperties>
            <includeProperty>^git.commit.id$</includeProperty>
          </includeProperties>
          <additionalProperties>
            <buildVersion>${project.version}</buildVersion>
          </additionalProperties>
          <verbose>true</verbose>
        </configuration>

Tell me if it's okay or if I need to change something.

Regards,

Marc

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,20 @@ It's really simple to setup this plugin; below is a sample pom that you may base
                       <!-- <excludeProperty>git.user.*</excludeProperty> -->
                     </excludeProperties>
 
+                    <!-- @since 2.1.14 -->
+                    <!--
+                        Can be used to include only certain properties into the resulting file.
+                        Will be overruled by the exclude properties.
+
+                        Each value may be globbing, that is, you can write {@code git.commit.user.*} to include both, the {@code name},
+                        as well as {@code email} properties into the resulting files.
+
+                        Please note that the strings here are Java regexes ({@code .*} is globbing, not plain {@code *}).
+                    -->
+                    <includeOnlyProperties>
+                      <!-- <includeOnlyProperty>^git.commit.id$</includeOnlyProperty> -->
+                    </includeOnlyProperties>
+
                     <!-- @since 2.1.10 -->
                     <!-- 
                       false is default here, if set to true it uses native `git` excutable for extracting all data.
@@ -605,6 +619,7 @@ Optional parameters:
 * **skip** - `(default: false)` *(available since v2.1.8)* - Skip the plugin execution completely.
 * **runOnlyOnce** - `(default: false)` *(available since v2.1.12)* - Use with caution! In a multi-module build, only run once. This means that the plugins effects will only execute once, for the parent project. This probably won't "do the right thing" if your project has more than one git repository. Important: If you're using `generateGitPropertiesFile`, setting `runOnlyOnce` will make the plugin only generate the file in the directory where you started your build :warning:. The `git.*` maven properties are available in all modules.
 * **excludeProperties** - `(default: empty)` *(available since v2.1.9)* - Allows to filter out properties that you *don't* want to expose. This feature was implemented in response to [this issue](https://github.com/ktoso/maven-git-commit-id-plugin/issues/91), so if you're curious about the use-case, check that issue.
+* **includeOnlyProperties** - `(default: empty)` *(available since v2.1.14)* - Allows to include only properties that you want to expose. This feature was implemented to avoid big exclude properties tag when we only want very few specific properties.
 * **useNativeGit** - `(default: false)` *(available since v2.1.10)* - Uses the native `git` binary instead of the custom `jgit` implementation shipped with this plugin to obtain all information. Although this should usualy give your build some performance boost, it may randomly break if you upgrade your git version and it decides to print information in a different format suddenly. As rule of thumb, keep using the default `jgit` implementation (keep this option set to `false`) until you notice performance problems within your build (usualy when you have *hundreds* of maven modules).
 * **abbrevLength** - `(default: 7)` *(available since v2.0.4)* - Configure the "git.commit.id.abbrev" property to be at least of length N (see gitDescribe abbrev for special case abbrev = 0).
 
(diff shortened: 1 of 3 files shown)


TheSnoozer:
Hi @Cousnouf,
thank you for your contribution and the time you have taken to implement the code and PR this into the project.

I'm always happy when I see PR's but I just want to check if I have understood the properties and new switches correctly...I'll just go through them...

1) includeProperties: this one seems to be obvious, but what's the benefit over the already existing  excludeProperties? By default everything is included and you can filter out everthing you do not want or you do not need. What should happen if you define a property in the exclude and include? As far as I could see in the code the exclude will win, but in this case it's a kind of confusion in the configuration of the includeProperties VS excludeProperties-Switch. However, exclude should definitely overrule the include. In my opionen everything would somehow works as expected but would confuse the user who need to configure the plugin and switches. 

2) additionalProperties: To be honest I didn't get the advantage of this switch. All properties defined in this List will just be added to the Project Properties. The Plugin itself does not use them afterwards or does not change them before adding them to the properties. Accoring to one comment in your files ``Can be used to add some properties in order to gather everything in the same file.`` it's only used to gather all information in one file....but why not using normal maven filter for that job?

Any further input on this is appreciated.

Thanks,

Cousnouf:
Hi Snoozer,

Thank you for your feedback, I appreciate it and it's interesting. I'll answer to your questions:

1) The benefits are obvious. In my case I just need one git property to be in the generated file. So, with include properties we go from there:

          <excludeProperties>
            <excludeProperty>git.*.user.*</excludeProperty><excludeProperty>git.*.message.*</excludeProperty>
            <excludeProperty>git.*.describe.*</excludeProperty><excludeProperty>git.*.branch.*</excludeProperty>
            <excludeProperty>git.*.remote.*</excludeProperty><excludeProperty>git.*.time.*</excludeProperty>
            <excludeProperty>git.dirty</excludeProperty><excludeProperty>git.tags</excludeProperty>
          </excludeProperties>

To there

    <includeProperties>
        <includeProperty>^git.commit.id$</includeProperty>
    </includeProperties>

I don't think there is any confusion because it's like many other options in several plugins where you can include files or exclude others. And then for exclude VS include I don't see anything complicated. You remove those you don't want or, if you only need a few, just include some. Or you can exclude one that is included in a glob pattern. I don't really think that it would be complicated for a developer to use or not use that property. 
Maybe we could rename it as includeOnlyProperties or something similar?

2) I'm quite okay with what you are saying. But I think it would be a complementary and useful option as it avoids to add another plugin or other tens of lines in the pom.xml only to concatenate another file that would contain the additional information. I see that maybe you don't want the commit id plugin to make everything and especially stuffs that are not directly linked. But that option seems just a small quick win for developers and would probably used. 
Or we could then just add the project.version property in the standard properties as it put the build time?

What do you think of that?

Regards,

Cousnouf:
Hi there, any updates please?

TheSnoozer:
Hi @Cousnouf ,
thanks for your detailed answers. I was on vacation and therefore didn't had the time to check for responses ;-)

I got the idea of the "includeProperties" however the name is still a bit confusing. I personally would suggest something like "includeOnlyProperties". Does this sound better or make it's duty more visible to the User? We definitely need to include this in the README. If you could please add this to the readme and outline that the includeOnlyProperties will definitely will be overruled by the exclude properties (it would be nice to have a test for that). It's had been a struggle in the past to add features without having a documentation for them..I now rather prefer to get the documentation with the PR :-P

For Number 2 I truly understand that it would make a developer's world easier however somehow this is not the purpose of the git-commit-id-plugin to add random properties to the project's properties. Sorry :-)
If you just need the build-time, this is already included with "git.build.time".


Thanks for your contribution

Cousnouf:
Hi @TheSnoozer,

I'm okay with the renaming of includePoperties to includeOnlyProperties, it will be more specific and precise.

For the second one I think you misunderstood the property I needed: as we already have the build time with git.build.time, could we also add the the maven build version? I'm okay with the fact we cannot put everything but the build version among the build time would be great. Especially as it in the same build context as the already present build time.

What do you think?

Marc

TheSnoozer:
Hi @Cousnouf ,
thanks for clarifying the new properties.
From my point it's ok when we including the project-version.
I'll review the PR today in the evening :-) 

Thanks for your contribution and fixing the minor thinks I have complained about ;)

Cousnouf:
Hello @TheSnoozer !

Thanks for the review. Keep me in touch if I have to change something.

Good feedback! It has ended to something more consistent I think!

TheSnoozer:
Hi @Cousnouf,
the code looks good for me.
Thanks for adding my feedback and thanks for your contribution.

Any other PRs are welcome :-)

Cousnouf:
Hi @TheSnoozer!

Thank you for all! Do you have any idea on when the 2.1.14 release is planned?

Cheers!

Marc

ktoso:
Releases happen pretty ad-hoc ;-) Think we've collected enough things to cut a release?
I could do it if so.

TheSnoozer:
I think a release would be nice to have.
We have collected a lot of version upgrades and some NPE fixes (a total of 59 commits): 
https://github.com/ktoso/maven-git-commit-id-plugin/compare/v2.1.13...master
I'll open a separate issue for the release :-)

Cousnouf:
Okay great! Thank you!

TheSnoozer:
@Cousnouf,
stay tuned for the release there are some minor problems @ktoso needs to fix before releasing the new version...

Release-Ticket can be found here https://github.com/ktoso/maven-git-commit-id-plugin/issues/193

Cousnouf:
Okay! I saw there are some encoding problems. If I can help let me know!

ktoso:
Finished my work day with 2 releases :-) Moved to a cafe and on to releasing this project then!

ktoso:
Released 2.1.14, [see milestone for changelog](https://github.com/ktoso/maven-git-commit-id-plugin/issues?q=milestone%3A2.1.14+is%3Aclosed)

Issue #54, opened 2012-12-05
Reveal describe details (tag name, number of commits, commit id) in their own properties


tuukkamustonen:
With describe returning something like 1.0.0-13-abc1234, we want each part in their separate property so we can build `finalName` with custom scheme.

Commit id is already put into separate field. Support would be needed for:

* Tag name
* Number of commits since last tag

Maybe as `git.closest.tag.name` and `git.closest.tag.commit.count` or something like that?

TheSnoozer:
Fixed by https://github.com/ktoso/maven-git-commit-id-plugin/pull/195

PR #195, opened 2015-05-24
#54: Reveal describe details (tag name, number of commits)

TheSnoozer:
Hi @ktoso,
today I had some time to implement the feature for #54.
This PR looks huge but it isn't.
Most what I did was shifting code from DescribeCommand.java to a new class JGitCommon.java to make it easier accessible. This basically gave me the opportunity to reuse the major part of the functionality. We still could speed up the JGit-Implementation by saving the values we need twice but I leave this as optional :-9

Also I fixed the license-headers (some files had a duplicated license info) and added the new properties to the readme (I caught one that was missing...).

Let me know if I should improve something for the PR.

Thanks,

### DIFF
--- a/src/main/java/pl/project13/jgit/JGitCommon.java
+++ b/src/main/java/pl/project13/jgit/JGitCommon.java
@@ -59,10 +59,6 @@ import static com.google.common.collect.Lists.newLinkedList;
 import static com.google.common.collect.Maps.newHashMap;
 import static com.google.common.collect.Sets.newHashSet;
 
-
-/**
- * @author <a href="mailto:konrad.malawski@java.pl">Konrad 'ktoso' Malawski</a>
- */
 public class JGitCommon {
   public Collection<String> getTags(Repository repo, final ObjectId headId) throws GitAPIException{
     RevWalk walk = null;
--- a/src/main/java/pl/project13/jgit/JGitCommon.java
+++ b/src/main/java/pl/project13/jgit/JGitCommon.java
@@ -1,0 +1,76 @@
+/*
+ * This file is part of git-commit-id-plugin by Konrad Malawski <konrad.malawski@java.pl>
+ *
+ * git-commit-id-plugin is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * git-commit-id-plugin is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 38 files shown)


ktoso:
Sounds great, reviewing!

src/main/java/pl/project13/jgit/JGitCommon.java
```diff
@@ -0,0 +1,367 @@
...
+import com.google.common.collect.Lists;
+
+import static com.google.common.collect.Lists.newArrayList;
+import static com.google.common.collect.Lists.newLinkedList;
+import static com.google.common.collect.Maps.newHashMap;
+import static com.google.common.collect.Sets.newHashSet;
+
+
+/**
+ * @author <a href="mailto:konrad.malawski@java.pl">Konrad 'ktoso' Malawski</a>
```
ktoso: Let's remove all `@author` tags by the way, to make nier to feel collective ownership about the code. I'll do that. :-)

src/main/java/pl/project13/jgit/dummy/DatedRevTag.java
```diff
@@ -1,3 +1,20 @@
...
+ * git-commit-id-plugin is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with git-commit-id-plugin.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
 package pl.project13.jgit.dummy;
```
ktoso: thanks

src/main/java/pl/project13/maven/git/JGitProvider.java
```diff
@@ -175,46 +193,40 @@ protected String getRemoteOriginUrl() throws MojoExecutionException {
...
-      });
-
-      Collection<String> tags = Collections2.transform(tagsForHeadCommit, new Function<Ref, String>() {
-        @Override public String apply(Ref input) {
-          return input.getName().replaceAll("refs/tags/", "");
-        }
-      });
-
+      ObjectId headId = headCommit.toObjectId();
+      Collection<String> tags = new JGitCommon().getTags(repo,headId);
```
ktoso: Why not reuse one instance? Is it stateful so it can't just be static methods?

ktoso: I'd opt for making it a constructor argument for each class needing it I guess.

TheSnoozer:
Well I think we could make it static.
In one of my projects I learned the hard way to avoid static method calls and rather go for new Clazz().method ;-)
I think this memory is burned into my head :-)

ktoso:
Or better just pass the instance around ;)

TheSnoozer:
Mhh I  think I would leave it as it is right now :-)
Passing only a single instance around or making the calls static is something I would leave optional for further improvement :-P

ktoso:
Though I would much rather have it passed in from the "beginning of the control flow" (the plugin, or the test) let's go with the current impl for now ;-)

ktoso:
Thanks, merged!

PR #197, opened 2015-05-30
fix for 196: create a uber-jar as a workaround for maven's class loading problem (MNG-5787)

TheSnoozer:
currently only need to include org.slf4j in the uber-jar

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -212,7 +212,7 @@
           <target>${java.target}</target>
         </configuration>
       </plugin>
-      <!-- we need to create a uber war for https://github.com/ktoso/maven-git-commit-id-plugin/issues/196 -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -212,6 +212,33 @@
           <target>${java.target}</target>
         </configuration>
       </plugin>
+      <!-- we need to create a uber war for https://github.com/ktoso/maven-git-commit-id-plugin/issues/196 -->
```
ktoso: uber jar :)

ktoso:
If it works LGTM, makes sense to jarjar it

pom.xml
TheSnoozer: fixed....think before typing :-)

ktoso:
Thanks!

TheSnoozer:
always welcome :-)

PR #200, opened 2015-06-03
Re enable Tests and update Readme for the latest version

TheSnoozer:


TheSnoozer:
I'll pull this in. Having an additional repro that is only for the single test is the best way to solve the failing test and makes it more clear what the test should really test. I needed to adjust that failing test once and tried it twice. However it's a matter of time until it will fail again, therefore I choosed to create a custom "git-environment" :-)

Issue #211, opened 2015-07-12
Maybe add a switch to generate the git.commit.id the "old" way


TheSnoozer:
I'm currently not really happy with the solution we have for the changing one of the major properties from ``git.commit.id`` to ``git.commit.id.full`` (related to https://github.com/ktoso/maven-git-commit-id-plugin/issues/122 and https://github.com/ktoso/maven-git-commit-id-plugin/commit/80fced8db322b1a1ccead59ca1ed8fed93742cec).

I think a way more better solution would be to include a switch (call that generateCommitIdOldFashioned) that's by default set to true.

switch set to true: will output everything as it was (so will generate the properties with ``git.commit.id``) and also will output an WARNING that using the switch set to true is deprecated.

switch set to false (manual adjustment): will output everything the new fashioned way (so will generate the properties with ``git.commit.id.full``) and no further WARNING/ERROR-Messages will be displayed to the user.

In this way, the user can choose himself whenever he would like to do the migration and we not just kill the backwards compatibility.

ktoso:
I agree the current style is not very nice, though not sure if we can afford breaking existing adjusted clients... will review thr PR

TheSnoozer:
fixed by https://github.com/ktoso/maven-git-commit-id-plugin/pull/212

PR #212, opened 2015-07-18
#211 : add a switch to generate the git.commit.id the "old" way

TheSnoozer:
As outlined in #211 I think this is much better for the generation "git.commit.id" VS "git.commit.id.full".
Have tested with different options and build in parallel (mvn -T 4 clean package -Pdemo -Dmaven.test.skip=true) nothing special discovered.

### DIFF
--- a/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
+++ b/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
@@ -404,16 +404,13 @@ public class GitCommitIdMojo extends AbstractMojo {
     }
 
     try {
-      if(commitIdGenerationMode == null){
-        commitIdGenerationMode = "flat";
-      }
-      switch(commitIdGenerationMode.toLowerCase()){
+      switch(CommitIdGenerationModeEnum.getValue(commitIdGenerationMode)){
       default:
         loggerBridge.warn("Detected wrong setting for 'commitIdGenerationMode' will fallback to default 'flat'-Mode!");
-      case "flat":
+      case FLAT:
         COMMIT_ID = COMMIT_ID_FLAT;
         break;
-      case "full":
+      case FULL:
         COMMIT_ID = COMMIT_ID_FULL;
         break;
       }
--- a/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
+++ b/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
@@ -336,9 +336,9 @@ public class GitCommitIdMojo extends AbstractMojo {
    * However, due to the fact that this is one of the major properties the plugin is exporting we just don't want to change the exporting mechanism and somehow throw the backwards compatibility away.
    * That's the point where this switch comes into place!
    * By default it is set to 'true' and will generate the formerly known property 'git.commit.id' as it was in the previous versions of the plugin. With keeping the switch set to 'true' the plugin will print a warning that using this switch set to 'true' is deprecated and may be removed in a future release. However keeping it to 'true' by default preserve backwards compatibility and allows to migrate to the new properties when it's convenient.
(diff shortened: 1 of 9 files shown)


src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
```diff
@@ -66,7 +66,11 @@
   // these properties will be exposed to maven
   public static final String BRANCH = "branch";
   public static final String DIRTY = "dirty";
-  public static final String COMMIT_ID = "commit.id.full";
+  @Deprecated
+  public static final String COMMIT_ID_OLD = "commit.id";
```
ktoso: It's not really deprecated since you/we plan to keep it around, right?

ktoso: i.e. the intention of this PR is to keep ir around

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
```diff
@@ -328,6 +332,21 @@
...
+   * By default it is set to 'true' and will generate the formerly known property 'git.commit.id' as it was in the previous versions of the plugin. With keeping the switch set to 'true' the plugin will print a warning that using this switch set to 'true' is deprecated and may be removed in a future release. However keeping it to 'true' by default preserve backwards compatibility and allows to migrate to the new properties when it's convenient.
+   * If you set this switch to 'false' the plugin will export the formerly known property 'git.commit.id' as 'git.commit.id.full'.
+   *
+   * *Note*: Depending on your plugin configuration you obviously can choose the 'prefix' of your properties by setting it accordingly in the plugin's configuration. As a result this is therefore only an illustration what the switch means when the 'prefix' is set to it's default value.
+   *
+   * @parameter default-value="true"
+   * @since 2.2.0
+   */
+  @Deprecated
+  private boolean generateCommitIdOldFashioned;
```
ktoso: I don't like the name here, it's not explaining what old and what new is. It should explain things about the nested-ness of the field.

src/main/java/pl/project13/maven/git/log/LoggerBridge.java
```diff
@@ -20,6 +20,7 @@
 public interface LoggerBridge {
   void log(Object... parts);
   void error(Object... parts);
+  void warn(Object... parts);
```
ktoso: +1

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
TheSnoozer: the intention of the PR is to keep the property around. However I think we should drop it somewhere in the feature and therefore I would say it's deprecated

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
TheSnoozer: Do you have something in mind how the name should be called?

ktoso: one of the modes seems to me like it could be called "flat", since it'a a flatt representation, and the other mode is "full", so maybe a configurable field instead of boolean flag, enabling different nesting modes?

TheSnoozer: I like that...will change it for the PR.
would call the field commitIdGenerationMode and options can be {'flat', 'full'}.
I think in case someone sets a wrong configuration it would be appropriate to throw an BuildException.

TheSnoozer:
Before pulling it into the repository I may need to change the "flat", "full" to a enum. Maven seem to support this (MNG-4292) with maven 3.0 onwards.

TheSnoozer:
I had finally choosen to go for a custom enum instead of rely on maven parameters....
Any feedback is welcome.

TheSnoozer:
I'll pull it in. In case there is something I should add let me know :-)

Issue #217, opened 2015-09-24
make generateGitPropertiesFilename optional


jmtd:
It would be great if generateGitPropertiesFilename was optional, defaulting to `src/main/resources/git.properties` (as the documented example).

ktoso:
Sounds like a good idea, thanks :)

ktoso:
Sounds like a good idea, thanks :)
If you have a bit of time to spare, a PR would be most wonderful, if not I hope we'll get to it eventually.

TheSnoozer:
Mhhh what's the benefit of defaulting to `src/main/resources/git.properties` instead of the current default `target/classes/git.properties` ?

the current default value is `${project.build.outputDirectory}/git.properties` is documented here:
https://github.com/ktoso/maven-git-commit-id-plugin/#configuration-options-in-depth

Or am I getting this wrong and something does not work as expected?

ktoso:
Now that you mention it, I think _I_ may have misunderstood hm... Maybe @jmtd means that in the `using the plugin` section, we're showing the `src/main...` value being used in this setting, and that's not the default, so it's unexpected, see:

```

                    <!-- The path for the to be generated properties file, it's relative to ${project.basedir} -->
                    <generateGitPropertiesFilename>src/main/resources/git.properties</generateGitPropertiesFilename>

```

Maybe then the solution is to document the right thing (I'm all for not changing existing behaviour in this case actually, don't want to break people's builds) - which would be `${project.build.outputDirectory}/git.properties` in that example (this line: https://github.com/ktoso/maven-git-commit-id-plugin/commit/c001e47c920a62de1eb39db5377226b0497c09d8#diff-04c6e90faac2675aa89e2176d2eec7d8R139 )

jmtd:
Hi,

> Now that you mention it, I think I may have misunderstood hm... Maybe @jmtd means that in the using the plugin section, we're showing the src/main... value being used in this setting, and that's not the default, so it's unexpected, see:

Thanks, yes - I didn't realise that the src/... value was *not* the default, and it wasn't clear whether the parameter could be committed. Please let me know what you think of #218.

TheSnoozer:
@jmtd,
Yes makes totally sense!

Merged your PR and will close this.
Thanks for reporting and fixing it :-)

ktoso:
Thanks guys!

PR #218, opened 2015-09-24
Correct default for generateGitPropertiesFilename

jmtd:
This is to address #217.


TheSnoozer:
Ahh I see. Makes sense now. Will merge it :-)

PR #225, opened 2015-11-19
properties: add git.dirty.mark property

paulmillar:
Motivation:

Easily include the dirty status in other maven properties, in the format
described by the 'dirty' configuration property.

Modification:

An additional property is exposed to maven: 'git.dirty.mark'.  If the
code-base has non-checked-in changes then the 'git.dirty.mark' property
value is the configured dirty mark, as used in 'git.commit.id.describe'.
If the code-base is clean then 'git.dirty.mark' expands to the empty
string.

Result:

Whether the code-base is clean or dirty is easily available in the
configured format.

### DIFF
--- a/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
+++ b/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
@@ -66,6 +66,7 @@ public class GitCommitIdMojo extends AbstractMojo {
   // these properties will be exposed to maven
   public static final String BRANCH = "branch";
   public static final String DIRTY = "dirty";
+  public static final String DIRTY_MARK = "dirty.mark";
   public static final String COMMIT_ID_FLAT = "commit.id";
   public static final String COMMIT_ID_FULL = "commit.id.full";
   public static String COMMIT_ID = COMMIT_ID_FLAT;
@@ -234,7 +235,7 @@ public class GitCommitIdMojo extends AbstractMojo {
    * As a general warning try to avoid three-letter time zone IDs because the same abbreviation are often used for multiple time zones.
    * Please review https://docs.oracle.com/javase/7/docs/api/java/util/TimeZone.html for more information on this issue.
    * will use the timezone that's shipped with java as a default (java.util.TimeZone.getDefault().getID())
-   * 
+   *
    * @parameter
    */
   @SuppressWarnings("UnusedDeclaration")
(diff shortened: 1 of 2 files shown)


TheSnoozer:
Hi @paulmillar, thanks for you contribution.
I highly appreachiate the time you have taken to implement this improvement.
However I don't see the difference to the ```git.dirty``` that's already exposed....
That's what you want hove here, right?

paulmillar:
Hi @TheSnoozer,

The property `git.dirty` is either `true` or `false`.
The property `git.dirty.mark` is either the empty string or the configured format for the dirty-mark.

It is possible to simulate `git.dirty.mark` value by (for example) creating a new property from 'git.dirty' with two string search-and-replace operations (e.g., reg-exp) to:
  replacing 'false' with the empty string,
  replacing 'true' with the formatted dirty-mark.

Such chained string manipulation is rather awkward to achieve in Maven.

It also violates the DRY principle[1].  If a project uses the git.commit.describe property but also uses the dirty-mark elsewhere then it must repeat the configuration of how the dirty mark should look.

This patch is also brings the dirty-mark into alignment with other properties.  The components of the git.commit.id.describe property are available separately (as git.commit.id.abbrev, git.commit.describe-short, git.closest.tag.name, git.closest.tag.commit.count).  The dirty-mark is the only component of git.commit.id.describe that is not available separately.

I hope that explains the motivation for the patch.

Cheers,

Paul.

[1] https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
```diff
@@ -66,6 +66,7 @@
   // these properties will be exposed to maven
   public static final String BRANCH = "branch";
   public static final String DIRTY = "dirty";
+  public static final String DIRTY_MARK = "dirty.mark";
```
ktoso: this breaks json output, I don't feel strongly enough about including this to break existing users.
If someone is using json output, they'd have:

```
dirty: false
```

And with your proposal they'd possibly break, since they could end up with:

```
dirty: { mark: ... }
```

so it's not a boolean anymore, and it's not possible to represent two values of `dirty` at the same time, thus we would end up breaking one or the other user.

ktoso:
I don't believe we can include this, esp not in a minor - see reasoning here: https://github.com/ktoso/maven-git-commit-id-plugin/pull/225/files#r57517684

Also, "it should not break people's code" trumps DRY at any given time in long term support of projects which thousands of others depend on.

ktoso:
Thanks for the idea, it's not a bad one, but sadly I don't think we can accept this PR.

Issue #228, opened 2015-12-01
Modularize plugin allowing lein / gradle / sbt usages of it as a library


ktoso:

TheSnoozer:
Done via https://github.com/git-commit-id/maven-git-commit-id-plugin/pull/441

PR #229, opened 2015-12-02
Move GitProviders into 'core' module.

mvitz:
These providers are reusable by other plugins.

To make this work they can't depend on any maven classes (e.g.
MojoExecutionException) so I had to introduce an own GitException.

Another change was to refactor the GitDataProvider mechanism from using
inheritance to composition. Therefore the GitProvider interface is now
the main abstraction used by the 'core' module. The GitDataProvider now
just assembles all properties by using a specific instance of the
GitProvider.

Relates to #228.


core/src/main/java/pl/project13/git/api/GitDescribeConfig.java
```diff
@@ -15,7 +15,7 @@
  * along with git-commit-id-plugin.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-package pl.project13.maven.git;
+package pl.project13.git.api;
```
ktoso: good package name choice :+1:

TheSnoozer:
Looks good for me. @ktoso Should I merge? :-)

ktoso:
Looks Good To Me! :-)
Let's merge to keep the momentum going!

Thanks a lot @mvitz for keeping at it!

TheSnoozer:
Note to myself: this was merged into `git-commit-id:modularization-effort` and never made it to the actual repo. Modularisation effort was done in https://github.com/git-commit-id/maven-git-commit-id-plugin/pull/441

PR #231, opened 2015-12-07
Fix for Slf4j misuse

bluestix:
This is a fix for: ktoso#230


ktoso:
This is now solved via https://github.com/ktoso/maven-git-commit-id-plugin/pull/233 I believe

PR #237, opened 2016-01-19
Improve docs

coladict:
By generating the file directly in the output directory, it will generally be ignored without changes to `.gitignore`.

Additionally, if not all the properties exist, that example class will explode on calling `toString()`


ktoso:
LGTM, thanks!

PR #245, opened 2016-04-29
Use `git ls-remote --get-url` to get the remote url

Shredder121:
Implementation detail and GOTCHA:
This also follows remote tracking branches, so if the name of the remote is different, then that is used.
image

I am usually on `origin`, but on other projects (like these) I am on `upstream`(since I can't push).
This means that the plugin doesn't fail when not used with `origin`.

Is this desired behavior?
If not, I can make it use `git ls-remote --get-url origin` instead.

Anyway, it does away with the parsing of porcelain output.
Let me know what you think.

### DIFF
--- a/src/main/java/pl/project13/maven/git/NativeGitProvider.java
+++ b/src/main/java/pl/project13/maven/git/NativeGitProvider.java
@@ -18,7 +18,6 @@ package pl.project13.maven.git;
 
 import static java.lang.String.format;
 
-import com.google.common.base.Splitter;
 import com.google.common.base.Throwables;
 import com.google.common.collect.Lists;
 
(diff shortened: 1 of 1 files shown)


Shredder121:
That was my second question, could someone test this? :sweat_smile:

Shredder121:
I'll see if I can also test it, using Windows here.

Shredder121:
So.
I can now guarantee that the input of `stripCredentialsFromOriginUrl` isn't null.
Do you want me to annotate that?

ktoso:
Yes please, the project tries to be consistent with use of those markers

Shredder121:
Hmm.. actually, only the `NativeGitProvider` now provides  nonnull.
The `JGitProvider` still provides nulls.

ktoso:
Good improvement as far as I can see, thanks!

Shredder121:
Nice, thank you!

Issue #260, opened 2016-09-02
git.properties not written if only the build time changes


nielsbasjes:
When creating the git.properties file I see this line after doing a mvn package
```
git.build.time=2016-09-02 @ 14\:04\:45 CEST
```
Now I run mvn package again and I see this in the output:
```
[INFO] git.build.time 2016-09-02 @ 14:06:57 CEST
```
followed by 
```
[INFO] Properties file [... /git.properties] is up-to-date (for module xxx)...
```
and as a consequence the git.properties file is not updated.

I want to make sure the git.properties file is update on every 'mvn package' even if nothing in the project has changed.

I had a quick look at the current code and found that apparently there is explicit code to make it work this way here https://github.com/ktoso/maven-git-commit-id-plugin/blob/master/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L616

Feature request: Add a flag I can set to always generate this file and get the correct build time in there.

TheSnoozer:
Hi,
understood the problem.
As far as I can remember the main reason why the "time" is not considered as a change is due to
https://github.com/ktoso/maven-git-commit-id-plugin/issues/151
and related issues. When introducing this switch, users of the plugin may be left alone with an endless build loop in Eclipse - not good at all ;-)


Why not use ``mvn **clean** package``?
Would including the native ``maven.build.timestamp`` be an option (https://dzone.com/articles/stamping-version-number-and)?

nielsbasjes:
I'm working on a web servlet, when I change the code and do restart from intellij I added this build time flag to my html so I can see the changed code actually was loaded.

Given this feature I (which I didn't know of) it sounds like the build time can be removed from this git plugin completely (with the addition of a bit of documentation).
I have to try it out to see what happens.

TheSnoozer:
I would add this specific case to the documentation instead of removing the build time from the plugin. Removing may break compatibility with users who read the property.

When using ``mvn clean package`` there should be no problem.

nielsbasjes:
Yesterday I found this rather 'evil' workaround https://github.com/nielsbasjes/yauaa/blob/master/webapp/pom.xml#L78

So I wrote a workaround to work around the workaround you wrote to avoid a bug in Eclipse ...

PR #261, opened 2016-09-05
Describe the subtle shift in meaning of git.build.time when generated into git.properties.

nielsbasjes:
Proposed solution for issue (only documentation) #260
Please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,10 @@ It's really simple to setup this plugin; below is a sample pom that you may base
                         which you can then normally read using new Properties().load(/**/)
                     -->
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -647,6 +650,20 @@ public GitRepositoryState(Properties properties)
 }
 ```
 
+Note on the generated git.build.time
+------------------------------------
+Note that when writing the git.porperties file the value *git.build.time* will only be updated when things in the commit information have changed.
```
ktoso: properties :)

ktoso:
Thanks for documenting this :)

PR #266, opened 2016-11-22
Add release notes support

pankajtandon:
Hi,
Thanks for the great work with this plugin.
I've created what I think is a good enhancement: Generation of release notes from git commit messages between tags.

The starting point is GitReleaseNotesMojo.
I've purposely added some description at the *end of* the readme so it breaks out the functionality. 

Please give it a spin and let me know how I can improve..
Thanks!
Pankaj


src/main/java/pl/project13/maven/git/GitDataProvider.java
```diff
@@ -98,10 +99,14 @@ public GitDataProvider setDateFormatTimeZone(String dateFormatTimeZone){
   protected abstract String getCommitMessageShort() throws GitCommitIdExecutionException;
   protected abstract String getCommitTime() throws GitCommitIdExecutionException;
   protected abstract String getRemoteOriginUrl() throws GitCommitIdExecutionException;
-  protected abstract String getTags() throws GitCommitIdExecutionException;
+  protected abstract String getTags()
```
ktoso: would have loved lo such reformattings :~ makes it hard to review

README.md
```diff
@@ -844,3 +844,81 @@ The best way to ask for features / improvements is [via the Issues section on gi
...
+A startTag is needed and then the plugin walks the commit graph to either:
+
+* Capture commit messages (that meet an optional regular expression) till the earliest commit
+* Capture commit messages (that meet an optional regular expression) till the endTag (specified in the configuration)
+
+#### Related Blog post
+
+A more detailed explaination can be found [here](http://technochord.com/2016/11/release-management-with-git)
+
+###Usage
```
ktoso: `### Usage`

README.md
```diff
@@ -844,3 +844,81 @@ The best way to ask for features / improvements is [via the Issues section on gi
...
+                        use this setting, otherwise we'll perform a search trying to
+                        figure out the right directory. It's better to add it explicitly IMHO.
+                    -->
+                    <dotGitDirectory>${project.basedir}/.git</dotGitDirectory>
+
+                    <!-- false is default here, it prints some more information during the build -->
+                    <verbose>false</verbose>
+                    
+                    <!-- The tag from which release notes need to be generated -->
+                    <startTag>release_3.1.4</startTag>
```
ktoso: hm, perhaps better put it into a group like

```
<releaseNotes>
  <startTag...
</releaseNotes>
```

README.md
```diff
@@ -844,3 +844,81 @@ The best way to ask for features / improvements is [via the Issues section on gi
...
+                    <!-- An optional regular expression that can be used to limit the commit messages that comprise
+                     the release notes. Often times developers will commit often but those commit messages may
+                     not have a relevance to the feature being developed.
+                     To prevent these "spurious" messages cluttering up the release notes, commits can either be squashed
+                        - OR -
+                     A tag like PRJ1-123 (for ticket 123 in project1 in the ticketing system) can be used in the commit 
+                     message. If such a convention is adopted, then the below regular expression will pick only those 
+                     messages to include in the release notes
+                     -->
+                    <commitMessageRegex>PRJ1-[0-9][0-9][0-9]</commitMessageRegex>
```
ktoso: this would also be in the releaseNotes tag then

src/main/java/pl/project13/maven/git/GitMojo.java
```diff
@@ -0,0 +1,326 @@
...
+import java.io.File;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.text.SimpleDateFormat;
+import java.util.List;
+import java.util.Properties;
+
+/**
+ * Created by pankaj on 10/21/16.
+ */
```
ktoso: please remove those - we follow [collective code ownership](http://www.extremeprogramming.org/rules/collective.html) and such created by don't help in that (feel free to remove any created by Konrad if there's any in there :-))

src/main/java/pl/project13/maven/git/GitReleaseNotesExecutionException.java
```diff
@@ -0,0 +1,26 @@
...
+        super(message, cause);
+    }
+
+    public GitReleaseNotesExecutionException(Throwable cause) {
+        super(cause);
+    }
+
+    public GitReleaseNotesExecutionException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace) {
+        super(message, cause, enableSuppression, writableStackTrace);
+    }
```
ktoso: if we only use one of the constructors (likely so) please just provide that one, remove the not used ones

src/main/java/pl/project13/maven/git/GitReleaseNotesMojo.java
```diff
@@ -0,0 +1,110 @@
...
+import java.io.FileOutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+
+/**
+ * This goal can be used to produce release notes between two tags.
+ *
+ * The release notes are comprised of all commit descriptions from the commit at which the first
+ * tag (release tag, possibily) exists and walks the commit graph,
+ * till the second (release) tag is reached.
```
ktoso: please show an example output in the README btw

src/main/java/pl/project13/maven/git/GitReleaseNotesMojo.java
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * Created by pankaj on 10/21/16.
+ */
+@Mojo(name = "releaseNotes", defaultPhase = LifecyclePhase.INITIALIZE, threadSafe = true)
+public class GitReleaseNotesMojo extends GitMojo {
```
ktoso: I'm also thinking if this would not be better served by being it's own plugin - WDYT?
Would you want to release it independently instead? Having a more focused impl could be beneficial (less cruft)

src/main/java/pl/project13/maven/git/GitReleaseNotesMojo.java
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * <p>The location of {@code 'release-notes.json'} file.
```
ktoso: won't this break on java8? It's pretty strict about formatting, please make it close tags

pankajtandon:
Hi @ktoso 
Thanks for the feedback. I will implement your suggestions.

But regarding your comment about breaking it out as a separate plugin:
There are a couple of reasons I think that this may be a good complementary functionality:

1. The audience is the same: Maven project stored in Git
2. The output is the same: A property file that shows git properties for the commitId goal and a json file to show release notes from the releaseNotes goal
3. The end user who is interested to expose git properties may also be interested in similarly exposing release notes.

But.. having said that, I do realize that it's a large refactor. 
Fortunately, you have decent test coverage so we can be fairly confident that nothing broke.

I just used a lot of the infrastructure code that you created like the LogBridge and the testing framework.
If I fork it, I'd really like to re-use a lot of code that you created.
Would that be ok?

Thanks a lot!

ktoso:
If you end up making it a separate plugin, I absolutely don't mind if you
reuse the infrastructure this plugin has built over the years - feel free
to use it, please adhere to the license this project is under :)

-- 
Konrad Malawski

pankajtandon:
Hi,
Trying to get the attention of @TheSnoozer :)

Can you please respond to @ktoso 's comments above?

TheSnoozer:
Sorry, was busy over Christmas xD

Well thats a though one.
I kinda like the idea of having release-notes generated.
I kinda dislike that it's currently not implemented for native git :-P

I'll look into the changes and see if it should be rather a separate plugin or should be integrated...

pankajtandon:
Thanks @TheSnoozer .. I'll await your feedback.
I will also address the comments that have been raised by @ktoso in an updated PR.

ktoso:
// Resolved conflicts.

TheSnoozer:
First of all sorry for the delay on this here and thanks for contributing to this project.
I finally had some time to think about this more and would agree with @ktoso and would say this rather fits a separate plugin.
Main reason being is that as a user of a git-plugin I might not want to generate release-notes for every build. This would be truly wasting resources for stuff that is simply not required to have it for *every* build and thus not being used at the end. Having an option to opt-out for the generation feels wrong and calls in for a separation of concerns (aka. different plugin with different life cycle). Furthermore I feel that there might be no one-fits-all solution for this use-case which would end up in adding a lot of code that kinda tries to patch the fact that multiple users may need to have different release-notes generated (e.g. release-notes for internal use VS release-notes you might hand out to customers). Next thing would be one would like to generate as HTML (user might want to provide a HTML template) or XML (user might want to provide a XSD), JSON (but needs to fit a specific formating), PDF (again a template issue), Word, Excel....this is way to complex to tackle all use-cases. Not to forget good old-style CSV release-notes.

Other reasons why I personally feel this might not fit in this project is that 
* a lot of open-source issue tracker already have integrated features to generate release-notes or in general I would assume ALL issue tracker have such features (if not open an issue with them)
* git somewhat offers a native way to generate release-notes - not sure if there would be something native for windows (e.g. ```git log v2.1.0...v2.1.1 --pretty=format:'<li> <a href="http://github.com/<user>/<project>/commit/%H">view commit &bull;</a> %s</li> ' --reverse | grep "#changelog"``` - see here https://coderwall.com/p/5cv5lg/generate-your-changelogs-with-git-log for more details or any other [shell-tricks guide](http://brettterpstra.com/2014/08/03/shell-tricks-changelogs-with-git/))
* Maven has a plugin for it (https://maven.apache.org/plugins/maven-changes-plugin/) that try to solve this problem
* Some git-projects (not limited to https://github.com/skywinder/Github-Changelog-Generator or https://github.com/vaab/gitchangelog) seem to try to tackle this problem
* There are plugins for CI-Tools like Jenkins available (not limited to https://wiki.jenkins.io/display/JENKINS/JIRA+Plugin or https://wiki.jenkins.io/display/JENKINS/Release+Helper+Plugin) that try to solve this problem

TLDR;
I truly like the Idea, but I don't see that it fits in this project.
Again thanks for your contribution and hope you either adjust this to fit your needs or find anything that would do what you need :-)

Issue #265, opened 2016-11-22
Feature request: Pretty-Print json


micw:
It would be cool, if the json would be more human readable (aka pretty printed). I use to have a version.json in my dists, that is read automatically - so it could be easily read by humans as well.

ktoso:
Should not be hard to implement, I'd suggest you'd try doing a PR which implements this - WDYT?
Happy to accept if such PR comes in.

micw:
Here you go. Sorry, my eclipse re-ordered the imports, the actual change is just one line.

ktoso:
Cool, thanks!

micw:
Thank you for merge. I'd be happy to see a new release with this feature ;)

PR #267, opened 2016-11-23
Use ObjectMapper's PrettyPrintWriter for JSON output.

micw:
Pretty-Print json #265


src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
```diff
@@ -634,7 +641,7 @@ void maybeGeneratePropertiesFile(@NotNull Properties localProperties, File base,
           if (isJsonFormat) {
             log.info("Writing json file to [{}] (for module {})...", gitPropsFile.getAbsolutePath(), project.getName());
             ObjectMapper mapper = new ObjectMapper();
-            mapper.writeValue(outputWriter, localProperties);
+            mapper.writerWithDefaultPrettyPrinter().writeValue(outputWriter, localProperties);
```
ktoso: yeah I guess that's a good idea in any case. LGTM

PR #271, opened 2016-12-30
Escaping unicode property values

mkrivan:
Using OutputStream instead of UTF-8 encoded Writer as Properties.store() first parameter the java properties file will be created in standard form (escaped unicode values like \u00F3 instead of ó).

Also I have added one line change when generating properties file also add to maven project properties (I think it was missing).


ktoso:
Would you mind adding a test that confirms this indeed fixes what it's supposed to?

There's `.git` directories which contain commits you could add one with the ó sign etc and run tests on it to see it writes properly.

mkrivan:
I have created test case as well as you see the commit log history. I have done some small commits for repairing some related issues while I was focusing on the properties file functionality. Al of them are correlated. Now the windows and linux behaviour is the same in both nativeGit and jGit case.

mkrivan:
Dear Konrad, 
Would you evaluate my suggestion? Many thanks.

ktoso:
Will try soon, battling some fires at work.
@TheSnoozer if you have time have a look too :)

src/test/java/pl/project13/maven/git/GitCommitIdMojoIntegrationTest.java
```diff
@@ -176,7 +176,12 @@ public void shouldGenerateCustomPropertiesFileProperties(boolean useNativeGit) t
     // given
     mavenSandbox.withParentProject("my-pom-project", "pom")
                 .withChildProject("my-jar-module", "jar")
-                .withGitRepoInChild(AvailableGitTestRepo.GIT_COMMIT_ID)
+                // I feel there could be a mistake in the AvailableGitTestRepo's enum values
+                // the MAVEN_GIT_COMMIT_ID_PLUGIN has the same directory as WITH_ONE_COMMIT_WITH_SPECIAL_CHARACTERS
+                // I think the GIT_COMMIT_ID has the value which should be planned to be the plugin repo itself
```
ktoso: please remove this comment, the value is OK, simply the same repo can fulfil different purposes AFAIR.

src/test/java/pl/project13/maven/git/GitPropertiesFileTest.java
```diff
@@ -0,0 +1,120 @@
...
+import static org.fest.assertions.Assertions.assertThat;
+import org.junit.runner.RunWith;
+import static org.mockito.internal.util.reflection.Whitebox.setInternalState;
+import pl.project13.maven.git.util.PropertyManager;
+
+@RunWith(JUnitParamsRunner.class)
+public class GitPropertiesFileTest extends GitIntegrationTest {
+
+  static final boolean UseJGit = false;
+  static final boolean UseNativeGit = true;
```
ktoso: Scala case style in Java app? ;) Please use Java style constants here. USE_JGIT, etc

src/test/java/pl/project13/maven/git/NativeAndJGitProviderTest.java
```diff
@@ -51,11 +51,20 @@
   public static final String DEFAULT_FORMAT_STRING  = "dd.MM.yyyy '@' HH:mm:ss z";
   public static final String ISO8601_FORMAT_STRING = "yyyy-MM-dd'T'HH:mm:ssZZ";
 
+// I feel there could be a mistake in the AvailableGitTestRepo's enum values
+// the MAVEN_GIT_COMMIT_ID_PLUGIN has the same directory as WITH_ONE_COMMIT_WITH_SPECIAL_CHARACTERS
+// I think the GIT_COMMIT_ID has the value which should be planned to be the plugin repo itself
+// If we use the GIT_COMMIT_ID repo this test will hang when nativeGit case selected in Windows definitely
```
ktoso: same comment as above? not sure this is needed, definitely should not stay in the codebase

src/test/java/pl/project13/maven/git/NativeAndJGitProviderTest.java
```diff
@@ -51,11 +51,20 @@
...
+// See the suggested modifications below
+
   @Test
   public void testCompareBasic() throws Exception
   {
     // Test on all available basic repos to ensure that the output is identical.
     for (AvailableGitTestRepo testRepo : AvailableGitTestRepo.values()) {
+      if (testRepo == AvailableGitTestRepo.GIT_COMMIT_ID) {
+        continue;
+      }
```
ktoso: please rewrite to not use continue, harder to track code.

mkrivan:
Many thanks for your review. I have modified based on your feedback.

ktoso:
Looks ok, thanks

PR #289, opened 2017-07-10
Update jgit to the latest version

candrews:
Update jgit to version 4.8.0.201706111038-r

It's always nice to use the latest version. All tests pass.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,7 +57,7 @@
     <maven-plugin-api.version>3.1.1</maven-plugin-api.version>
     <maven-plugin-plugin.version>3.4</maven-plugin-plugin.version>
 
-    <jgit.version>4.3.1.201605051710-r</jgit.version>
+    <jgit.version>4.5.2.201704071617-r</jgit.version>
     <junit.version>4.12</junit.version>
     <mockito.version>2.0.35-beta</mockito.version>
 


TheSnoozer:
Thanks for the PR!
I love to see some libraries getting updated...

This seems to drop Java 1.7...at least the test on travis fails with java 1.7

```
ERROR] Failed to execute goal org.apache.maven.plugins:maven-plugin-plugin:3.4:descriptor (default-descriptor) on project git-commit-id-plugin: The API of the mojo scanner is not compatible with this plugin version. Please check the plugin dependencies configured in the POM and ensure the versions match. org/eclipse/jgit/api/GitCommand : Unsupported major.minor version 52.0 -> [Help 1]
```

Even though Java SE 7 doesn't get any further Public Updates (since Apr 2015) it is still supported
- with Premier Support until Jul 2019
- with Extended Support until  Jul 2022

Need to look into this further before I'll pull this in....not sure if creating a branch with this would be smarter...

candrews:
JGit 4.5.x is the last version to support Java 7. I've updated this PR to use that - sorry I didn't notice the Java 8 failure before.

TheSnoozer:
Thanks for the quit change! Tests are all green now :-)
Just for reference...found the ticket where it was changed in jgit (https://bugs.eclipse.org/bugs/show_bug.cgi?id=500059   -- Target Milestone: v4.6)

Issue #310, opened 2017-08-25
git-commit-id-plugin removes project properties generated by other plugins


cbuschka:
We love the git-commit-id-plugin and use it in combination with the jacoco plugin. The jacoco plugin creates a property which must be included in surefire runs as argLine.

When property filtering is active in the git-commit-id-plugin and generating a properties file is disabled, the git-commit-id-plugin operates on the maven project properties and removes all project properties not matching the patterns of <includeOnlyProperties>, including the jacoco generated property and others not created by the git-commit-id-plugin.

Workaround: Disable property filtering.

I created an example setup to reproduce the issue at https://github.com/cbuschka/maven-git-commit-id-plugin-issue310.

TheSnoozer:
see https://github.com/ktoso/maven-git-commit-id-plugin/pull/311

TheSnoozer:
Thanks again for reporting and fixing this!
Was merged with #311

Issue #138, opened 2014-11-23
git.branch character replacer


taorg:
If you work with git-flow, branches get names like this "feature/feature_name"
It would be nice to create jar o ear names with finalNames like this <finalName>${epack}-${git.commit.id.describe}-${git.commit.time}-${git.branch}</finalName>

But the caracter "/" is not valid as filename, it would be nice to have the possibility to replace this character.

troinine:
We have similar issues with other fields thus it would be nice to be able to filter any of the git properties, not just `git.branch`. We use the commit message field and therefore someone can easily end up putting some special characters like double quotes which will fail in our case (Spring Boot app with YAML properties).

TheSnoozer:
From a plugin perspective...what comes close to this feature is:
https://code.google.com/archive/p/maven-replacer-plugin/

The Build Helper Mojo also has something similar:
http://www.mojohaus.org/build-helper-maven-plugin/regex-property-mojo.html#replacement
However I would separate the concerns of validation and replacement...

It might be even useful to restrict the replacement to a specific property:
```
<replacement>
    <property>git.branch</property>
    <token>(?:[^\/]|\/\/)+</token>
    <value>$1-$2</value>
</replacement>
```

Not sure if it would be smarter to include another Mojo or execute it inside the git-commit-id-plugin.
However there should be something like ```failOnUnableToReplacePropertyInfo``

TheSnoozer:
I have implemented this feature and it will be available in version 2.2.3.

Example:
```
<replacementProperties>
  <replacementProperty>
    <property>git.branch</property>
    <token>^([^\/]*)\/([^\/]*)$</token>
    <value>$1-$2</value>
  </replacementProperty>
</replacementProperties>
```

Please checkout the readme for additional details!

berniefitz82:
I'm having a similar issue where local machines git.branch returns just the branch name but during CI/CD it returns origin/branch-name

TheSnoozer:
please create a new issue with all relevant details.

PR #311, opened 2017-08-25
#310: Only properties created by maven-git-commit-id-plugin filtered.

cbuschka:
Added extra checks if a property has been generated by maven-git-commit-id-plugin before filtering. This fixes issue 310.


TheSnoozer:
Hi thank you for reporting the issue and creating a pull request!
This is indeed kinda ugly....I'm just thinking if there would be other things effected when using the outlined scenario...for example in the latest version of the plugin has an option to replace properties...however using the outlined scenario the replacement of properties would work differently and thus the question raises why we even use different property sets inside initProperties() see
https://github.com/ktoso/maven-git-commit-id-plugin/blame/master/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L538.

The PR looks ok, but I'll definitely need to look into that further to check if other items of the plugin would be affected by this as well. Also would be nice to have a testcase for this since it is something more critical ;-)

cbuschka:
Yes you are right I will create a unit test for.

cbuschka:
Ok, I moved the filtering into a collaborator (PropertiesFilterer) and added a unit test.

TheSnoozer:
Thanks for the test cases!
Yes, fully agree! The `properties = initProperties();` should rather be always `properties = new Properties()`.

It seems that the `project.getProperties().putAll(properties);` here https://github.com/ktoso/maven-git-commit-id-plugin/blob/master/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L383 is independent from the `generateGitPropertiesFile` switch.
Currently the initProperties get the mavenProject-properties when `generateGitPropertiesFile=false`
and if `generateGitPropertiesFile=true` the properties are being attached with the `project.getProperties().putAll(properties);` outlined above.

Please add the refactoring and I'll pull in the changes....
Thanks again for reporting and creating a PR :-)

cbuschka:
Moved replacement code into PropertiesReplacer with own test and added the refactoring described above. The refactoring changes behaviour! What is the intend behind the properties replacement? Should it be performed on project wide properties?

TheSnoozer:
Actually its a really good question on what properties the replacement should perform. The initial intent comes from here https://github.com/ktoso/maven-git-commit-id-plugin/issues/138. In a nutshell a branch has a name like "feature/feature_name" and someone wanted to use the branch-name in the finalName of the artifact. However the character "/" is not valid as filename and thus this would not work.

Based on what I wrote inside the docs (https://github.com/ktoso/maven-git-commit-id-plugin/commit/1e8b523cddf0d44ca664e1e158fa80f0866acb4f) the replacement can be used to replace certain characters or strings using regular expressions within the generated git properties. I even added the following note `Please note that the replacement will *only be applied to properties that are being generated by the plugin.`...

I would say at the time being it is sufficient to have the replacement only work on the properties generated by the plugin (that is pretty much what the readme says about the functionality).

Just for reference:
We either can adjust the way it works (it requested with a string reason) or the user can use 
https://code.google.com/archive/p/maven-replacer-plugin/ which is pretty much the same but for all properties.

TLDR
======
All changes look good to me!
Thanks for the code cleanup!
I will pull in the changes since the replacement should currently only work on properties generated by the plugin (was not released yet, so no one can claim it ever worked differently :-P)

cbuschka:
Thanks for merging! I am glad I could contribute and give something back in return.

ktoso:
Thanks @cbuschka !

This is now released as 2.2.3, have fun! :)

PR #312, opened 2017-08-28
Add version badge

ktoso:


ktoso:
Add version badge so it's nicely visible what version is the latest 
via https://github.com/jirutka/maven-badges

Issue #315, opened 2017-08-30
maven.gitcommitid.skip does not override configuration in POM


jgerken:
Executing Maven with the property `-Dmaven.gitcommitid.skip=true` does not skip the plugin execution when the configuration in the POM includes a `<skip>false</skip>`.  Thus, the command line option does not override the setting in the POM.

Usually, all properties set via command line override the settings in the POM (this is even the purpose of the properties), e.g. skipping the tests or skipping linting. In my opinion, the behaviour of the git-commit-plugin should be changed to follow the general Maven approach and let command line property override the setting in the POM.

For now, an easy workaround exists, simply remove `<skip>false</skip>` from the POM.

TheSnoozer:
We are using plain maven features and not doing any other magic (here is where the parameter comes from: https://github.com/ktoso/maven-git-commit-id-plugin/blob/master/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L218)

I'll investigate but fear that this is a maven bug. What maven version are you using?

TheSnoozer:
Congratulation you found what seems to be a Maven-Bug!
See Sample-Project:
[SampleProject.zip](https://github.com/ktoso/maven-git-commit-id-plugin/files/1265346/SampleProject.zip)

> mvn clean install && mvn clean initialize -PdemoConfigSet -Dmaven.buildHelperMojo.skip=true
    doesn't work since it has the configuration-tag set

> mvn clean install && mvn clean initialize -PdemoConfigUnSet -Dmaven.buildHelperMojo.skip=true
    works as expected since it has the configuration-tag NOT set

Reported here:
https://issues.apache.org/jira/browse/MNG-6278

TheSnoozer:
Just saying:
The maven-surefire-plugin is using the same mechanism.
Not have tested this but a possible workaround would be using the same mechanism as [maven-surefire-plugin](http://maven.apache.org/plugins-archives/maven-surefire-plugin-2.12.4/examples/skipping-test.html).
They use as configuration inside the configuration tag the value `<skipTests>` and via command line `maven.test.skip` and if you check closely the source code it actually two distinct parameters:
* via configuration `<skipTests>` see https://github.com/apache/maven-surefire/blob/master/maven-surefire-common/src/main/java/org/apache/maven/plugin/surefire/AbstractSurefireMojo.java#L160
[...]

PR #319, opened 2017-09-09
#315: deploying the workaround used inside maven-surefire-plugin to be able to skip via configuration and commandline properly

TheSnoozer:


ktoso:
Refs #315

TheSnoozer:
IK it's inside to force the developers of plugins to to such bad things....but thats how it is. Thats how it is deployed inside the maven surfire-plugin....

Issue #316, opened 2017-09-07
Don't require Maven patch version


cbismuth:
Hi,

Is really useful to depend on Maven 3.1.1 or would it be enough to depend on 3.1.0 and let the end-user patch his own installation?

Thanks,
Christophe

TheSnoozer:
Hi,
I might not understand the full need of this....

Regardless to answer your question:
The plugin need to depend on some Maven version and this version can not be set by the user.
It furthermore can not be set by the user, since its set inside the [pom of the plugin](https://github.com/ktoso/maven-git-commit-id-plugin/blob/master/pom.xml#L57) and required for building the plugin.

Please checkout the [README](https://github.com/ktoso/maven-git-commit-id-plugin/#plugin-compatibility-with-maven) to see what maven version are required to run the plugin. If you still rely on an old 3.1.0 version I'd recommend the plugin version `2.1.13`, or if possible to upgrade your maven version to `3.1.1` or anything more recent.

cbismuth:
Hi,

Yes we have an outdated CI installation. I think there might not be any required API change or blocking bug on your side, so why not depend on 3.1.1 and not 3.1.0?

I'll upgrade our CI dependencies, but this question is more about good practice in plugin development: why not depend on the lowest required version of Maven?

TheSnoozer:
> I think there might not be any required API change or blocking bug on your side, so why not depend on 3.1.1 and not 3.1.0?
Knowing Maven I wouldn't assume this - you never know ;-)

Regardless as a Plugin Developer I would argument the other way round:
Why can't you update your maven-version to anything more recent? If you can use a maven version that was released on 2013-10-04 (about 4 years ago) what system are you using? Is this still supported by anything?
Usually with updates more features are being shipped or security patches fixed and thus I would rather tend and say why not upgrade to the latest and greatest? 

Regardless of this I kinda understand where you are coming from and I can check locally if I see an impact if the version changes, but knowing Maven there is 1209831092 trillion things that are breaking with this simple version change. Furthermore I would also argue why not go back even further and support the 3.0 version then. When going back to 3.0 I would assume even more breaks but I'll check on this...

TheSnoozer:
Just as a personal note:
[...]

PR #322, opened 2017-09-16
Support Maven 3.X

TheSnoozer:
This Refs to https://github.com/ktoso/maven-git-commit-id-plugin/issues/316
Local tests with
```
#!/bin/bash
for mavenBasePath in /usr/share/maven-3*
do
  echo "============================================"
  ${mavenBasePath}/bin/mvn --version
  echo ""
  # ${file}/bin/mvn clean initialize -Pdemo
  for pluginVersion in 2.2.4-SNAPSHOT
  do
    rm -rf target
    # echo ${pluginVersion}
    echo "${mavenBasePath}/bin/mvn pl.project13.maven:git-commit-id-plugin:${pluginVersion}:revision"
    ${mavenBasePath}/bin/mvn pl.project13.maven:git-commit-id-plugin:${pluginVersion}:revision -Pdemo > /dev/null
    rc=$?;
    if [[ $rc != 0 ]]; then
      echo "pl.project13.maven:git-commit-id-plugin:${pluginVersion} is NOT compatible";
    else
      echo "pl.project13.maven:git-commit-id-plugin:${pluginVersion} is compatible"
    fi

    echo ""
    cat target/testing.properties | sed 's/\(.*\)/  \1/'
  done
done
```
didn't yield any execution problems.
However for older Maven-Versions (3.0.X) there is a warning from SLF4J:
```
[INFO] --- git-commit-id-plugin:2.2.4-SNAPSHOT:revision (get-the-git-infos) @ git-commit-id-plugin ---
[INFO] dotGitDirectory /home/stefan/workspace/maven-git-commit-id-plugin/.git
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```
This does not seem to impact the execution of the plugin and thus will be a known issue.

Also this updates the travis-build in a way that
- java 7, 8, 9 is being tested (as previously)
- plugin is being build with maven versions 3.0, 3.1.1, 3.2.5, 3.3.9 and 3.5.0

I would not add tests for any other minor versions to keep the builds to a somewhat minimum.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,30 @@ jdk:
   - oraclejdk9
   - oraclejdk8
   - openjdk7
+
+matrix:
+  include:
+  - jdk: openjdk7
+    env: CUSTOM_MVN_VERION="3.0"
+  - jdk: openjdk7
+    env: CUSTOM_MVN_VERION="3.1.1"
+  - jdk: openjdk7
+    env: CUSTOM_MVN_VERION="3.2.5"
+  - jdk: openjdk7
+    env: CUSTOM_MVN_VERION="3.3.9"
+  - jdk: oraclejdk9
+    env: CUSTOM_MVN_VERION="3.5.0"
+
+install:
+  - if [[ -n "${CUSTOM_MVN_VERION}" ]]; then
+      echo "Download Maven ${CUSTOM_MVN_VERION}....";
+      if [[ "${CUSTOM_MVN_VERION}" == "3.0" ]]; then
+        wget https://archive.apache.org/dist/maven/binaries/apache-maven-3.0-bin.zip || travis_terminate 1;
+      else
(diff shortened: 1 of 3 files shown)


TheSnoozer:
Thinking more about what I just pushed.
The different builds with Maven might not be what I want to test with travis.
i want to verify if the build plugin can be executed within travis under certain maven projects.
Currently it only tests if the project can be build with different maven versions (which should always be the case).

Tests would needed to be adjusted to reflect that.
Don't merge this yet.

TheSnoozer:
Just as an interesting side-note:
I did some testing and everything seems to behave as expected/wanted.
I just stumbled over one think where I was wondering why the build passed with ```<maven-plugin-api.version>3.1.1</maven-plugin-api.version>``` and maven 3.0. This should technically exactly what this thing is doing. Well blame maven for it and it was fixed for 3.0.2 (see https://issues.apache.org/jira/browse/MNG-4840).

PR #314, opened 2017-08-29
Modularize v3

TheSnoozer:
I know we tried this before...just thinking to bring this back alive ;P

tried to use the existing branch but it seems to be easier just to redo it since a lot of stuff has changed over time...

PR #323, opened 2017-09-17
Update branch modularization effort

TheSnoozer:
Alright....as outlined in https://github.com/ktoso/maven-git-commit-id-plugin/pull/314 I took the time to merge origin/master into modularization-effort....

There are some things I saw I feel we would need to cleanup a bit, but in general it looks okayish and the build is passing (now also running out of time to tidy up even more...still want to have a weekend xD)

In general this is what I feel doesn't seem to be right:
* [GitDataProvider](https://github.com/TheSnoozer/maven-git-commit-id-plugin/blob/modularization-effort/maven/src/main/java/pl/project13/maven/git/GitDataProvider.java#L44) and [AbstractBaseGitProvider](https://github.com/TheSnoozer/maven-git-commit-id-plugin/blob/modularization-effort/core/src/main/java/pl/project13/git/impl/AbstractBaseGitProvider.java#L37) share some common variables like abbrevLength and gitDescribe. The solution I currently implemented kinda ships around this issue but this simply looks wrong.
* The Maven-Module currently has a dependency to [`org.eclipse.jgit` for scope `test`](https://github.com/TheSnoozer/maven-git-commit-id-plugin/blob/modularization-effort/maven/pom.xml#L125) which is being used for tests inside maven-jgit-package [(e.g. DescribeCommandAbbrevIntegrationTest)](https://github.com/TheSnoozer/maven-git-commit-id-plugin/blob/modularization-effort/maven/src/test/java/pl/project13/maven/jgit/DescribeCommandAbbrevIntegrationTest.java#L22). Some of the tests also use a dependency to MavenProject so some of them can't be moved straight forward. The easiest solution I could come up with, without rebuilding the entire test-set is introduce this dependency which is not that bad because it would only affect test-scope (still would need to be cleaned up)

Thats pretty much all of the really ugly stuff I still have in mind when merging.


TheSnoozer:
refs https://github.com/ktoso/maven-git-commit-id-plugin/issues/228

TheSnoozer:
Updated:
220335d..1a7f8a7  modularization-effort -> modularization-effort

TheSnoozer:
Super seeded by https://github.com/git-commit-id/maven-git-commit-id-plugin/pull/441.

PR #345, opened 2017-12-04
Add commit count 'rev-list HEAD --count'

velo:


TheSnoozer:
Hi,
thanks for your contribution!

I'd like to provide some feedback on this:
- [ ] I needed to actually look at the [documentation](https://www.kernel.org/pub/software/scm/git/docs/git-rev-list.html) what this particular rev-list command does since the name is doesn't speak for itself. I'd suggest to rename it to something more meaningful like ``git.commit.count`` or  ``git.commit.total.count``
- [ ]  This project uses checkstyle to ensure that the codestyle is somewhat consistent for formating code. The breaking travis checks complains that your code currently do not follow those code style requirements. Please check them by running ``mvn clean verify -Pcheckstyle -Dmaven.test.skip=true -B``. If you are using java9 you may want to run ``mvn clean verify -Pcheckstyle -Dcheckstyle.version=8.2 -Dmaven.test.skip=true -B``
- [ ] Please add this new property to the readme (or otherwise no one knows about it)
- [ ] I saw that the native git command executes ``runQuietGitCommand(canonical, "rev-list HEAD --count");`` and the jgit uses ``RevWalkUtils.count(revWalk, evalCommit, null); ``. This would yield different results for jgit VS native git executable if someone sets the plugin configuration ``evaluateOnCommit`` to something that is not HEAD. In a nutshell this configuration allows to tell the plugin what commit should be used as reference to generate the properties from.
The native git implementation would need to be changed to ``runQuietGitCommand(canonical, "rev-list --count " + evaluateOnCommit);`` since execution like ``git rev-list --count HEAD^1`` works with the native git as well.
- [ ] No tests have been introduced (at least we should check for some cases if this property is generated correctly...and that the output jgit VS git is the same)

Please don't get this wrong. New ideas are very welcome here, just need to make sure the code-base stays in a maintainable condition.

Thanks,

ktoso:
With added tests by @TheSnoozer in https://github.com/ktoso/maven-git-commit-id-plugin/pull/357
Thanks, we'll release the feature eventually / hopefully soon

Issue #355, opened 2018-01-23
Bamboo generates unsupported gitRemoteString


DanielKuehner:
We are using the Bamboo build server to build our maven projects. 

### Expected Behavior
The build should continue to succeed after adding this plugin.

### Current Behavior
With this plugin the build fails.

### Steps to Reproduce
Maven prints the following error:
`[ERROR] Failed to execute goal pl.project13.maven:git-commit-id-plugin:2.2.4:revision (default) on project provided: Could not complete Mojo execution... java.net.URISyntaxException: Illegal character in authority at index 7: file://D:\[DIRECTORY]\_git-repositories-cache\c05bf5b346c29887ddbfdd9911f13aee6eeb63bb -> [Help 1]`

This is because we are using the "Enable repository caching on remote agents" option for our git repository inside the Bamboo configuration. This produces the following entry within the `.git/config`:
`[remote "origin"]`
`	url = file://D:\\[DIRECTORY]\\_git-repositories-cache\\c05bf5b346c29887ddbfdd9911f13aee6eeb63bb`

If we disable the caching option the build succeeds. Because the origin url points to `file://nothing`.

According to #258 this is an unsupported gitRemoteString and one should open another ticket for other formats. Is it possible to add this format too? Because we cannot alter the Bamboo behavior.

### Environment
- Plugin version 2.2.4
- Java version 1.8.0_162
- Atlassian Bamboo version 5.15.7 build 51521
- Apache Maven 3.5.2 (138edd61fd100ec658bfa2d307c43b76940a5d7d; 2017-10-18T09:58:13+02:00)
- Windows 7
- Maven execution: mvn.cmd --batch-mode -Djava.io.tmpdir=C:\Users\[USER]\AppData\Local\Temp\SOME_DIR package

TheSnoozer:
Hi,
thanks for reporting your issue here.

This different file handling and different prefixes are a pain in the a**.

[...]

PR #356, opened 2018-01-23
#355: Bamboo generates unsupported gitRemoteString with file://[DIR]

TheSnoozer:
see https://github.com/ktoso/maven-git-commit-id-plugin/issues/355 for reference.


TheSnoozer:
Missing change:
- instead of failing, just log an error and return an empty remote url (saver fall back instead of failing the build)

PR #363, opened 2018-03-13
Read the output of native git before waiting for it to complete

Calixte:
The current NativeGitProvider waits for the git process to be finished before it starts reading the output. This can cause a deadlock on some systems where the buffer size is smaller than the git output, for example when there are a lot of git tags.

Here is the documentation of [Process](https://docs.oracle.com/javase/7/docs/api/java/lang/Process.html) for reference:
> The parent process uses these streams to feed input to and get output from the subprocess. Because some native platforms only provide limited buffer size for standard input and output streams, failure to promptly write the input stream or read the output stream of the subprocess may cause the subprocess to block, or even deadlock.


ktoso:
Thanks!

PR #369, opened 2018-06-24
add build number support (git.build.number & git.build.number.unique)

### LINKED ISSUE #353: Include jenkins BUILD_NUMBER inside git properties
I am using this plugin for track my changes. But only the commit id is not enough. I want to add the version number in jenkins which is set in env to the git properties.

secustor:
added support for CI build numbers. 

I have tested this locally and for Gitlab CI on a server, but I have no Jenkins or Travis server around. 

This should fix #353


README.md
```diff
@@ -780,6 +788,17 @@ If you need the actual *build time* then simply use the a filtered properties fi
 git.build.time=${git.build.time}
 ```
 
+Note on the git.build.number module
```
ktoso: It's not a module though?

README.md
```diff
@@ -780,6 +788,17 @@ If you need the actual *build time* then simply use the a filtered properties fi
...
+ ------------------------------------
+ Be aware that *git.build.number* is a project specific build number and *git.build.number.unique* contains the server wide
```
ktoso: Rather reword this to explain what it is and where it is available please.
E.g.: 

```
The `git.build.number` variables are available on some hosted CIs and can be used to identify the "number" of the build. Refer to the table below to see which values are supported by which CIs
```

Sounds better?

Note also that the name should be wrapped in ` and not *.

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
```diff
@@ -525,6 +526,41 @@ void loadGitData(@NotNull Properties properties) throws GitCommitIdExecutionExce
...
+   *  build.number.unique is a server wide unique build number
+   * @param properties a properties instance to put the entries on
+   */
+  void loadBuildNumber(@NotNull Properties properties) {
+    String buildNumber = null;
+    String uniqueBuildNumber = null;
+
+    //JENKINS
+    if (System.getenv().containsKey("BUILD_NUMBER"))  {
+      buildNumber = System.getenv().get("BUILD_NUMBER");
```
ktoso: does this work? If yes, then please include this info in the readme -- it did not mention that jenkins also works

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
```diff
@@ -525,6 +526,41 @@ void loadGitData(@NotNull Properties properties) throws GitCommitIdExecutionExce
     }
   }
 
+  /**
+   * Fill the properties file build number environment variables which are supplied by build servers
+   *  build.number is the project specific build number, if this number is not available the unique number will be used
```
ktoso: weird formatting? Remove 1 space

ktoso:
I'm ok with merging this even though not really a "git thing"

README.md
secustor: K, I will change that. Would "Note on git.build.number variables" work for you?

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
TheSnoozer: IMHO the GitCommitIdMojo is already messy as it is, and doesn't need extra logic. Why not move this into the GitDataProvider? 
It has quite similar code for retrieving all sorts of environment variables if we run on a CI server.
Should we always try to extract the properties or only if `runningOnBuildServer` is true?

secustor: In that case I suggest that all code related to `git.build..*` variables will be moved to  a new class (e.g. `BuildDataMojo.java`).         
From there we can then detect the kind of the build enviroment and not only if the plugin is running on a CI server.       
In my opinion this would clean up the code and would allow further, if wanted, extensions to the build variables.

@TheSnoozer  My idea was that somebody might want to run a script to set the buid number manually.  And the Jenkins enviroment variable ("BUILD_NUMBER") is pretty straight forward for this.

TheSnoozer:
Just wondering about the status, shall I review the latest changes here?
The last commit indicates that this is "WIP"....I usually ignore everything marked as WIP :-P

secustor:
@TheSnoozer I have not come yet to test it througly and I'm still investigating how I can read out properties from Bamboo, but the structure will not change much anymore.     
If you think you know a cool feature/varialbe just say it. :wink: 

I will add documentation and make requested changes on weekend.

TheSnoozer:
Ok, thanks for doing this :-)
Absolutely no rush, just wanted to check :-)
I'll give it a review when home.

secustor:
@TheSnoozer I have implemented the support TeamCity. You can review it now.

src/main/java/pl/project13/maven/git/build/TeamCityBuildServerData.java
```diff
@@ -0,0 +1,60 @@
...
+import pl.project13.maven.git.log.LoggerBridge;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.util.Map;
+import java.util.Properties;
+
+public class TeamCityBuildServerData extends BuildServerDataProvider {
+
+  private final Properties systemProperties = new Properties();
```
TheSnoozer: Calling this variable ``systemProperties`` is slightly confusing since the naming creates the illusion it contains some properties from the system package.
Please rename it to ``teamcityBuildProperties``.

src/main/java/pl/project13/maven/git/build/TeamCityBuildServerData.java
```diff
@@ -0,0 +1,60 @@
...
+
+  TeamCityBuildServerData(LoggerBridge log) {
+    this(log, System.getenv());
```
TheSnoozer: Mhh I don't like that we here again fetch the system environment variables within the constructor.

Why not initialize the teamcityBuildProperties with null like so:
```private Properties systemProperties = null;```
Whenever it is needed within the functions use ``getTeamcityBuildProperties(env)`` with:
```java
private Properties getTeamcityBuildProperties(@NotNull Map<String, String> env) {
    if (teamcityBuildProperties == null) {
      try {
        teamcityBuildProperties = new Properties();
        teamcityBuildProperties.load(new FileInputStream(env.get("TEAMCITY_BUILD_PROPERTIES_FILE")));
      } catch (IOException e) {
        log.error("Failed to retrieve Teamcity properties file", e);
      }
    }
    return teamcityBuildProperties;
}
```

Could the statement ``new FileInputStream(env.get("TEAMCITY_BUILD_PROPERTIES_FILE"))`` potentially throw a NullPointerException if ``env.get("TEAMCITY_BUILD_PROPERTIES_FILE")`` returns null for some reason??

src/main/java/pl/project13/maven/git/build/BuildServerDataProvider.java
```diff
@@ -0,0 +1,151 @@
...
+  public abstract BuildEnvironmentType getBuildEnvironmentType();
+
+  /**
+   * Get the {@link BuildServerDataProvider} impelementation for the running environment
+   * @param env environment variables which get used to identify the environment
+   * @param log logging provider which will be used to log events
+   * @return the corresponding {@link BuildServerDataProvider} for your environment or {@link UnknownBuildServerData}
+   */
+  public static BuildServerDataProvider getBuildServerProvider(@NotNull Map<String, String> env, LoggerBridge log) {
+    switch (checkBuildEnvironmentForType(env)) {
```
TheSnoozer: I believe with with one of my previous comments this would be the last reference of the ``checkBuildEnvironmentForType`` method call.

The switch and checks done in this specific function somewhat look duplicated to what is being done here.
In any case we somewhat return an instance of a BuildServerDataProvider, why not get rid of the ``checkBuildEnvironmentForType`` method and change it to:
```java
public static BuildServerDataProvider getBuildServerProvider(@NotNull Map<String, String> env, LoggerBridge log) {
    if (BambooBuildServerData.isActiveServer(env)) {
      return new BambooBuildServerData(log);
    }
    if (GitlabBuildServerData.isActiveServer(env)) {
      return new GitlabBuildServerData(log);
    }
.....
}
```

It seems that with all the rework the enum ``BuildEnvironmentType`` would also become not needed anymore (I think we still need to the distinction between an unknown environment and and actual build server, but this could be done with ``if(buildServerDataProvider instanceof UnknownBuildServerData)``

src/main/java/pl/project13/maven/git/GitDataProvider.java
```diff
@@ -167,57 +169,19 @@ void validateAbbrevLength(int abbrevLength) throws GitCommitIdExecutionException
    * @param env environment settings
    * @return results of getBranchName() or, if in Jenkins/Hudson, value of GIT_BRANCH
    */
-  protected String determineBranchName(Map<String, String> env) throws GitCommitIdExecutionException {
-    if (runningOnBuildServer(env)) {
-      return determineBranchNameOnBuildServer(env);
+  protected String determineBranchName(@NotNull Map<String, String> env) throws GitCommitIdExecutionException {
+    if (!BuildServerDataProvider.checkBuildEnvironmentForType(env).equals(BuildEnvironmentType.UNKNOWN)) {
+      String branchName = BuildServerDataProvider.getBuildServerProvider(env,log).getBuildBranch(env,log);
```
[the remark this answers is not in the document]
secustor: I thought that this way it would be easier to use it externaly if needed, but I see your point.

src/main/java/pl/project13/maven/git/build/BuildServerDataProvider.java
```diff
@@ -0,0 +1,151 @@
...
+
+public abstract class BuildServerDataProvider {
+
+  private final LoggerBridge log;
+  private String dateFormat = "yyyy-MM-dd'T'HH:mm:ssZ";
+  private String dateFormatTimeZone = null;
+  private String prefixDot = "";
+  private MavenProject project = null;
+
+  BuildServerDataProvider(LoggerBridge log) {
```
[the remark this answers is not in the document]
secustor: sure.

src/main/java/pl/project13/maven/git/build/BuildServerDataProvider.java
secustor: Yes you are right we can drop the `BuildEnvironmentType` and the corresponding methods too.

TheSnoozer:
Thanks for all the changes and especially taking care of this MR!
Code looks good to me now.

Will merge even though it might not be a "git thingy"  ;-)

Issue #350, opened 2017-12-07
Can we replace Guava with Java 8 features? If we keep it should be upgraded to a more recent version


TheSnoozer:
Just a thought....we use guava for some common utility stuff.
With Java 8 some of Guava's stuff can be replaced with native Java code.
If there would no additional benefit in using guava it should be a replaced with native Java code.
This is just an evaluation if we can replace all current features with native code.
If not we keep guava (but should upgrade it to a more recent version).

Only thing that we seem to use (and might be useful) is an immutable map where [no such replacement seem to exist inside the native java 8 environment](https://softwareengineering.stackexchange.com/questions/221762/why-doesnt-java-8-include-immutable-collections).

https://www.reddit.com/r/java/comments/217rnf/what_is_obsolete_in_guava_since_java_8/
> FluentIterable
> Stream
> 
> {Listenable,Settable}Future
> CompletableFuture
> 
> Sets.newConcurrentHashSet()
> ConcurrentHashMap.newKeySet()
> 
> Joiner.on(separator).join(...)
> String.join(separator, ...)
> ...or new StringJoiner(separator).add(...)
> ...or stream.collect(Collectors.joining(separator))
> 
> com.google.common.base.{Predicate,Function,Supplier,Optional}
> java.util.function.{Predicate,Function,Supplier}, java.util.Optional (and Optional{Int,Long,Double})
> 
> Ordering.natural()
> Comparator.naturalOrder()
> 
> ordering.reverse()
> comparator.reversed()
> 
> Ordering.compound(comparator1, comparator2)
> comparator1.thenComparing(comparator2)
> 
> Ordering.from(comparator).onResultOf(function)
> Comparator.comparing(function, comparator)
> 
> Ordering.from(comparator).nulls{First,Last}()
> Comparator.nulls{First,Last}(comparator)
> 
> ComparisonChain
> comparator.thenComparing(...)
> 
> Some uses of BaseEncoding
> java.util.Base64
> 
> Lists.transform(list, function)
> list.stream().map(function).collect(Collectors.toList())
>  - More verbose, less efficient.
> 
> Maps.uniqueIndex(collection, function)
> collection.stream().collect(Collectors.toMap(Function.identity(), function))
>  - More verbose, not immutable.
> 
> Multimaps.index(collection, function)
> collection.stream().collect(Collectors.groupingBy(function))
>  - More verbose, not immutable, not a Multimap.

[...]

Issue #346, opened 2017-12-04
Drop support for Java 7?


TheSnoozer:
Just a general question:
Would it be safe to drop the support for Java 7?
Who is using this now rather old Java (hopefully nearly everyone has upgraded to at least java 8)?

Based on [Oracle Java SE Support Roadmap](http://www.oracle.com/technetwork/java/eol-135779.html) Java 7 does not get any Public Updates, Premier Support ends Jul 2019; Extended Support ends Jul 2022. 

| Release | GA Date | Premier Support Until | Extended Support Until | Sustaining Support |
| ------- | --------- | ---------------------- | ------------------------ | ------------------- |
| 6 | Dec 2006 | Dec 2015 | Dec 2018 | Indefinite |
| 7 | Jul 2011 | Jul 2019 | Jul 2022 | Indefinite |
| 8 | Mar 2014 | Mar 2022 | Mar 2025 | Indefinite |
| 9 (non-LTS) | Sep 2017 | Mar 2018 | Not Available | Indefinite |

Considering it is "only" covered by Premier Support for quite a while now I'm debating to just stop supporting it...maybe still to early and it should wait another year...

Main question to consider the upgrade to a new Java version:
How much is this project gaining by dropping the support?
- The main thing to consider is that this upgrade would allow the project to use more recent versions of third party libraries (e.g. jgit v4.6 moved towards 1.8 only https://bugs.eclipse.org/bugs/show_bug.cgi?id=500059)
- I wouldn't expect crazy changes for the resources inside the project itself since I personally only see lambda expressions one key factor for java 8 and the rest is rather a more convenient call to some java functionality http://www.oracle.com/technetwork/java/javase/8-whats-new-2157071.html). However I don't see a particular thing where lambda expression would fit into this project.

Support for Java 8 and Java 9 is not debatable and should remain fully supported.
Note: Maven (https://maven.apache.org/docs/history.html) currently still supports Java 7

ktoso:
Yes, we absolutely should drop 7 already I think.
People who are on 7 are free to remain on an old version of the plugin, as they do with their entire runtime - being on an unsupported version.

Feel free to change the build's lang level etc

TheSnoozer:
sounds good, will drop it for next release :-)

ktoso:
Yay :)

laurentperez:
+1
[...]

Issue #336, opened 2017-11-01
Plugin hangs when too many changes in the git-repo


andreas-eternach:
The plugin execution blocks when there are too many changes in the git-repo.

Probably because the stdout from git-child-process is processed after git-process terminates.

In case stdout-buffer is full the git-child-process stops working and a deadlock occurs.

**Details:** 
- Plugin-Version:2.2.3
- git-command for child process "git status -s"
- Stack traces from maven environmen

`2017-11-01 08:23:18
Full thread dump Java HotSpot(TM) 64-Bit Server VM (25.144-b01 mixed mode):
"RMI TCP Connection(2)-192.168.99.1" #16 daemon prio=5 os_prio=0 tid=0x00000000223f7000 nid=0x2e08 runnable [0x000000002430e000]
java.lang.Thread.State: RUNNABLE
at java.net.SocketInputStream.socketRead0(Native Method)
at java.net.SocketInputStream.socketRead(SocketInputStream.java:116)
at java.net.SocketInputStream.read(SocketInputStream.java:171)
at java.net.SocketInputStream.read(SocketInputStream.java:141)
at java.io.BufferedInputStream.fill(BufferedInputStream.java:246)
at java.io.BufferedInputStream.read(BufferedInputStream.java:265)
locked <0x000000076e07b818> (a java.io.BufferedInputStream)
at java.io.FilterInputStream.read(FilterInputStream.java:83)
at sun.rmi.transport.tcp.TCPTransport.handleMessages(TCPTransport.java:550)
at sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.run0(TCPTransport.java:826)
at sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.lambda$run$0(TCPTransport.java:683)
at sun.rmi.transport.tcp.TCPTransport$ConnectionHandler$$Lambda$6/2140250730.run(Unknown Source)
at java.security.AccessController.doPrivileged(Native Method)
at sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.run(TCPTransport.java:682)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
at java.lang.Thread.run(Thread.java:748)
Locked ownable synchronizers:
[...]

PR #382, opened 2018-09-01
#346 / #336 / #350: Drop Java 1.7 support / introduce a timeout for any native git command / replace almost every guava feature with Java 8 features

TheSnoozer:
This changing with this MR:
- #346: Drop support for Java 1.7
- #336: Introduce a new configuration to allow to set a timeout for the native binary call's
- #350: update guava and replace almost every guava feature with Java 8 features
- Update dependencies and some cleanup
- Updated issue templates, created pull request template


pom.xml
```diff
@@ -107,9 +107,9 @@
 
     <!-- IntelliJ Annotations -->
     <dependency>
-      <groupId>com.intellij</groupId>
+      <groupId>org.jetbrains</groupId>
       <artifactId>annotations</artifactId>
-      <version>12.0</version>
+      <version>16.0.2</version>
     </dependency>
```
TheSnoozer: Alternatively we could use
```
<dependency>
		<groupId>com.google.code.findbugs</groupId>
		 <artifactId>jsr305</artifactId>
		 <version>3.0.2</version>
	</dependency>
```
However, technically I don't see a huge benefit in changing this...the findbugs one might look more `native` java like...

src/main/java/pl/project13/jgit/JGitCommon.java
```diff
@@ -63,38 +61,32 @@ public JGitCommon(LoggerBridge log) {
...
+                final RevCommit objectCommit = finalWalk.parseCommit(objectId);
+                if (finalWalk.isMergedInto(objectCommit, tagCommit)) {
+                  return true;
+                }
+              } catch (Exception ignored) {
+                log.debug("Failed while getTags [{}] -- ", tagRef, ignored);
+              }
+              return false;
+            })
+            .map(tagRef -> tagRef.getName().replaceAll("refs/tags/", ""))
```
TheSnoozer: Why not use ``trimFullTagName(tagRef.getName())`` here?

This specific Method would use ``replaceFirst("refs/tags/", "")`` instead of ``replaceAll``. However AFAIK we are expecting only the start to be ``refs/tags/...``. I think a ``replaceAll`` would prevent us from using actual tags that are looking like ``refs/tags/...`` (very unlikely that this ever happens). In practice I wouldn't assume it actually makes a difference.

Issue #366, opened 2018-04-27
Eclipse (maybe M2E-WTP) +git-commit-id-plugin reliable interaction


lrozenblyum:
### Expected Behavior
When I create a new commit, I expect that maven-git-commit-id-plugin will update git.properties in the target classes 

### Current Behavior
At the moment git.properties file is not updated reliably when working in Eclipse and creating new commits (NOTE: in standalone Maven it works really fine).
Maven -> Update project could help.

### Steps to Reproduce (for bugs only)
<plugin>
			    <groupId>pl.project13.maven</groupId>
			    <artifactId>git-commit-id-plugin</artifactId>
			    <version>2.2.4</version>
			    <executions>
			        <execution>
			            <id>get-the-git-infos</id>
			            <goals>
			                <goal>revision</goal>
			            </goals>
			        </execution>
			    </executions>
			    <configuration>
			        <dotGitDirectory>${project.basedir}/.git</dotGitDirectory>
			        <prefix>git</prefix>
			        <verbose>false</verbose>
			        <generateGitPropertiesFile>true</generateGitPropertiesFile>
			        <generateGitPropertiesFilename>${project.build.outputDirectory}/git.properties</generateGitPropertiesFilename>
			        
			        <format>properties</format>
			        
			        <includeOnlyProperties>git.commit.id.abbrev</includeOnlyProperties>
			    </configuration>
			</plugin>

### Environment (for bugs only)
Eclipse Oxygen.3a
Oracle JDK1.6.0.45
Maven 3.3.9 (built-in into Eclipse)
Windows7
Maven is executed automatically by M2E (and possible M2E-WTP since it's a web project).

NOTE: I also tried in-memory solution (when the plug-in doesn't generate a separate file) and it also work in Eclipse not perfectly.

Thank you for any help and suggestions

lrozenblyum:
In principle I agree that's not a huge problem.
However it is (can) be a problem when we do switch to a branch and the plug-in doesn't recognize need to generate git.properties and thus we have a crash in the web application which depends on it.

TheSnoozer:
Fixed by https://github.com/ktoso/maven-git-commit-id-plugin/pull/385

Issue #269, opened 2016-12-14
M2E Lifecycle Mapping default <runOnIncremental>true not a good idea


vorburger:
Hello, thanks for this cool Maven plugin!

It's great and very useful - but slows down workspace builds in Eclipse.. scanning non-trivial git repos has a price, and I'm not sure doing that at every incremental build is such a good idea.. how about changing the [runOnIncremental in lifecycle-mapping-metadata.xml](https://github.com/ktoso/maven-git-commit-id-plugin/blob/master/src/main/resources/META-INF/m2e/lifecycle-mapping-metadata.xml#L11) from its current default true to false?

It's of course possible to change this in a pom.xml in org.eclipse.m2e:lifecycle-mapping, or Eclipse ws/.metadata/.plugins/org.eclipse.m2e.core/lifecycle-mapping-metadata.xml - what I'm suggesting here is that the current default is perhaps sub-optimal ... if you do agree to change it, anyone who truly does want it to run in-IDE incrementally all the time, and who's willing to take the workspace build performance hit for it, can still re-enable it, just like I'm currently ignore it.

Would you welcome a contribution for this?

TheSnoozer:
Hi,
thanks for your valuable feedback!
Based on https://wiki.eclipse.org/M2E_compatible_maven_plugins the runOnIncremental is the following:
> If <runOnIncremental/> is set to true, corresponding mojo will be executed during both full and incremental workspace builds. If set to false (the default), the mojo will be executed during full workspace build only. 

Currently I think the maven-git-commit-id-plugin should be executed with full and incremental builds as well. My reasoning behind this is that with my favorite environment the full build will only be triggered when asked for. Thus let's just imagine I have a clean project and trigger a full build. This will then trigger the maven-git-commit-id-plugin and generate the git properties accordingly. However now I find a bug, fix it and commit it. With the runOnIncremental set to true the git properties will be updated to the correct new commit version. However when using the runOnIncremental set to false the  git properties won't be updated since no full build has been. This will result in outdated information. Not sure if someone really relies on the properties that had been generated by Eclipse / Idea but I kinda feel this should be as is......

Also when talking about "the plugin is slow" - have you tried the native git implementation? :-)

michaldo:
I agree with @vorburger that runOnIncremental should be false. maven-git-commit-id-plugin  is similar to maven-javadoc-plugin. As application developer I don't care my javadocs are outdated. I also don't care git.properties are outdated.
[...]

PR #385, opened 2018-09-20
run on eclipse incremental build if properties file was not generated

PPazderski:
This may fix issue #366 while still respecting performance concerns from #269.

It use a BuildContext as described in [m2e](https://www.eclipse.org/m2e/documentation/m2e-making-maven-plugins-compat.html#buildcontext) to distinguish between incremental and normal (e.g. mvn command line) builds. On incremental builds it should only run if the git.properties file is missing at all.

Additional the BuildContext can inform eclipse about the new generated file. (especially useful if file is generated in some folder not ignored by the IDE)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -76,6 +76,11 @@
       <artifactId>maven-core</artifactId>
       <version>${maven-plugin-api.version}</version>
     </dependency>
+    <dependency>
+     <groupId>org.sonatype.plexus</groupId>
+     <artifactId>plexus-build-api</artifactId>
+     <version>0.0.7</version>
+    </dependency>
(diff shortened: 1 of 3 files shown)


TheSnoozer:
Thanks for your contribution!
I'll need some time to read the referenced documentation and would like to perform some testing.

When merging this with `master` this will become available with some 3.X version. Would that be sufficient for you? Or would you prefer to have it also available in some 2.X version?

PPazderski:
I do not need it in 2.x.

Since our project is quite small I enabled property update on every incremental build through workspace lifecycle-mappings.

TheSnoozer:
Thanks. Looks good for me.

PR #386, opened 2018-09-25
Support jgit 5.1.1, add direct httpclient, and update dependencies

hazendaz:
### Context

Support jgit 5.1.1 release.  Currently this plugin is using httpclient which was a transient dependency of jgit.  That transient is no longer present.  Therefore, for continued usage, a simple change to include the httpclient along with jgit 5.1.1 works without issue.  Additionally updated mockito/junit/jackson to latest releases.  All changes isolated to pom only.

### Contributor Checklist
- [ N/A] Added relevant integration or unit tests to verify the changes
- [ N/A] Update the Readme or any other documentation (including relevant Javadoc)
- [ X] Ensured that tests pass locally: `mvn clean package`
- [ N/A] Ensured that the code meets the current `checkstyle` coding style definition: `mvn clean verify -Pcheckstyle -Dmaven.test.skip=true -B`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,9 +57,9 @@
     <maven-plugin-api.version>3.0</maven-plugin-api.version>
     <maven-plugin-plugin.version>3.5</maven-plugin-plugin.version>
 
-    <junit.version>5.2.0</junit.version>
     <mockito.version>2.21.0</mockito.version>
     <jgit.version>5.1.1.201809181055-r</jgit.version>
+    <junit.version>5.3.1</junit.version>
 
     <fest-assert.version>1.4</fest-assert.version>
   </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -88,7 +88,7 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-databind</artifactId>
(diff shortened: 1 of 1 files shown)


TheSnoozer:
Hi, thanks for the contribution!
Wow those dependencies got outdated fast!

Do we actually need the httpclient anywhere in this project?
If not, I'd suggest to just remove it completely.

I'm also a bit surprised that this removal didn't resulted in a major version bump in the jgit version number....but oh well you noticed it :-)

hazendaz:
@TheSnoozer Yes, httpclient is used.  I discovered that today as I had been probably for some really old reason been overriding jgit in this plugin.  After upgrading it crashed due to httpclient usage.  It's using URIBuilder from that within the project currently.

TheSnoozer:
Ahh I see. Thanks for the clarification.
It's [currently be used](https://github.com/ktoso/maven-git-commit-id-plugin/blob/63bfac7cdf2ac142877f02dc12e777d5d88eafb8/src/main/java/pl/project13/maven/git/GitDataProvider.java#L254) for making sure that the password inside the remote git (http) url is hidden when forwarded to the properties.

Thanks again :-)

PR #388, opened 2018-09-26
fixes #387 'runOnlyOnce' & --project make Mojo not run at all

### LINKED ISSUE #387: 'runOnlyOnce' coupled with Maven's --projects param makes plugin not run at all
### Describe the bug
In the case of a multi-module Maven project, when we use the 'runOnlyOnce=true' parameter coupled with 'injectAllReactorProjects=true', we expect the plugin to run once to calculate all git variables and we expect those variables to be available in children projects.

But this is not the case when we run Maven on a specific sub-project:
` mvn install --projects mySubProject`

neither when we use this command to resume from a specific sub-project:
` mvn install -rf mySubProject`

### Steps to Reproduce
- Here is the config in the parent pom:

```xml
			<plugin>
				<version>2.2.5</version>
				<executions>
					<execution>
						<id>get-the-git-infos</id>
						<goals>
							<goal>revision</goal>
						</goals>
						<phase>initialize</phase>
						<configuration>
							<generateGitPropertiesFile>false</generateGitPropertiesFile>
							<runOnlyOnce>true</runOnlyOnce>
							<injectAllReactorProjects>true</injectAllReactorProjects>
							<verbose>true</verbose>
							<skipPoms>false</skipPoms>
						</configuration>
					</execution>
				</executions>
			</plugin>
```

### Expected behavior

` mvn install --projects mySubProject`

we expect to see the verbose logs of all the variables calculated.
But in fact we get this:

 [INFO] --- git-commit-id-plugin:2.2.5:revision (get-the-git-infos) @ mySubProject ---
 [INFO] runOnlyOnce is enabled and this project is not the top level project, skipping execution!

### Additional context

version tested is the last released: 2.2.5

### My analysis:

The code in GitCommitMojo looks like 
```java
if (runOnlyOnce) {
        if (!session.getExecutionRootDirectory().equals(session.getCurrentProject().getBasedir().getAbsolutePath())) {
          log.info("runOnlyOnce is enabled and this project is not the top level project, skipping execution!");
          return;
        }
      }
```

=> so the code expects to run at least once from Maven's ExecutionRootDirectory. But there it does not happen.

My suggestion is to set a variable (user variable ?) the first time the plugin is ran and when this variable is already set, we bypass the execution in case 'runOnlyOnce=true' 
I will provide a Pull Request soon.

fmarot:
This commit fixes
https://github.com/ktoso/maven-git-commit-id-plugin/issues/387

### Contributor Checklist
- [ ] Ensured that tests pass locally: `mvn clean package`

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,7 +313,8 @@ It's really simple to setup this plugin; below is a sample pom that you may base
                     <!--
                        Use with caution!
 
-                       In a multi-module build, only run once. This means that the plugins effects will only execute once, for the parent project.
+                       In a multi-module build, only run once. This means that the plugins effects will only execute once, even if the
+                       parent projet is not in the reactor (when you use Maven's --project option or --resume-from ).
                        This probably won't "do the right thing" if your project has more than one git repository.
 
                        Important: If you're using `generateGitPropertiesFile`, setting `runOnlyOnce` will make the plugin
--- a/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
+++ b/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
@@ -63,7 +63,10 @@ import javax.annotation.Nullable;
  */
 @Mojo(name = "revision", defaultPhase = LifecyclePhase.INITIALIZE, threadSafe = true)
 public class GitCommitIdMojo extends AbstractMojo {
-  /**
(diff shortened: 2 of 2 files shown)


TheSnoozer:
Closing based on the better proposed solutions in #387.

Issue #24, opened 2012-03-27
git.branch returning wrong information


ctmay4:
I am seeing different results when running under Linux and Windows.  When I run the plugin through Windows during development, I get the correct git.branch value:

```
git.commit.id.abbrev=5a3086c
git.commit.user.email=mayc@imsweb.com
git.commit.message.full=Branch doesn't seem to be working, so removing the method to get it.\n
git.commit.id=5a3086cfbc994014179d2c573ced4150d143cc0e
git.commit.message.short=Branch doesn't seem to be working, so removing the method to get it.
git.commit.user.name=Chuck May
git.build.user.name=Chuck May
git.build.user.email=mayc@imsweb.com
git.branch=master
git.commit.time=27.03.2012 @ 17\:22\:02 EDT
git.build.time=27.03.2012 @ 17\:38\:48 EDT
```

However when the project is build through continuous integration (Jenkins running under Linux), the git.branch is incorrect:

```
git.commit.id.abbrev=5a3086c
git.commit.user.email=mayc@imsweb.com
git.commit.message.full=Branch doesn't seem to be working, so removing the method to get it.\n
git.commit.id=5a3086cfbc994014179d2c573ced4150d143cc0e
git.commit.message.short=Branch doesn't seem to be working, so removing the method to get it.
git.commit.user.name=Chuck May
git.build.user.name=Unknown
git.build.user.email=Unknown
git.branch=5a3086cfbc994014179d2c573ced4150d143cc0e
git.commit.time=27.03.2012 @ 17\:22\:02 EDT
git.build.time=27.03.2012 @ 17\:26\:52 EDT
```

My pom.xml looks like your sample:

```
            <plugin>
                <groupId>pl.project13.maven</groupId>
                <artifactId>git-commit-id-plugin</artifactId>
                <version>2.0.3</version>
                <executions>
                    <execution>
                        <goals>
                            <goal>revision</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <verbose>true</verbose>
                    <dotGitDirectory>${project.basedir}/.git</dotGitDirectory>
                    <generateGitPropertiesFile>true</generateGitPropertiesFile>
                    <generateGitPropertiesFilename>src/main/resources/git.properties</generateGitPropertiesFilename>
                </configuration>
            </plugin>
```

antofran:
Hi

[...]

PR #393, opened 2018-10-31
Issue #24 Add property for using CI branch name or not

jonevn:
### Context
Added property for setting if branch name should be taken from build environment or not.
When using JGitflow maven plugin, then the property from Jenkins can not be used, so this will
open up for use with that plugin. The property will be true by default, so that old behavior is used.

### Contributor Checklist
- [Could not run with maven, but executed in Eclipse] Ensured that tests pass locally: `mvn clean package`


TheSnoozer:
Hi,
thanks for the MR.
The changes look good to me.

Travis seem to fail due to the usage of an unsupported oraclejdk (https://travis-ci.community/t/help-with-oraclejdk10-build-on-linux/398/8).

I'll just merge it...

PR #396, opened 2018-11-05
#336: fix hanging git subprocess in case of too much output on stdout/stderr.

andreas-eternach:
### Context
#336: fix hanging git subprocess in case of too much output on stdout/stderr.

The previous implementation read the output after the subprocess finished. In case the buffer is full, the subprocess waits infinitly.

Therefore the new implementation flushes the subprocess buffer by reading from stdout/stderr on background threads.

### Contributor Checklist

### DIFF
--- a/src/main/java/pl/project13/maven/git/NativeGitProvider.java
+++ b/src/main/java/pl/project13/maven/git/NativeGitProvider.java
@@ -27,7 +27,6 @@ import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
 import java.util.concurrent.TimeUnit;
 
-
 public class NativeGitProvider extends GitDataProvider {
 
   private transient ProcessRunner runner;
@@ -236,7 +235,7 @@ public class NativeGitProvider extends GitDataProvider {
   public String getRemoteOriginUrl() throws GitCommitIdExecutionException {
     return getOriginRemote(canonical, nativeGitTimeoutInMs);
   }
-  
+
   @Override
   public String getClosestTagName() throws GitCommitIdExecutionException {
     try {
@@ -291,14 +290,12 @@ public class NativeGitProvider extends GitDataProvider {
   /**
    * Runs a maven command and returns {@code true} if output was non empty.
    * Can be used to short cut reading output from command when we know it may be a rather long one.
-   *
    * Return true if the result is empty.
-   *
    **/
   private boolean tryCheckEmptyRunGitCommand(File directory, long nativeGitTimeoutInMs, String gitCommand) {
     try {
       String env = System.getenv("GIT_PATH");
-      String exec = (env == null) ? "git" : env;
+      String exec = env == null ? "git" : env;
       String command = String.format("%s %s", exec, gitCommand);
 
       return getRunner().runEmpty(directory, nativeGitTimeoutInMs, command);
@@ -311,7 +308,7 @@ public class NativeGitProvider extends GitDataProvider {
 
   private String runQuietGitCommand(File directory, long nativeGitTimeoutInMs, String gitCommand) {
     final String env = System.getenv("GIT_PATH");
-    final String exec = (env == null) ? "git" : env;
+    final String exec = env == null ? "git" : env;
     final String command = String.format("%s %s", exec, gitCommand);
 
     try {
@@ -323,7 +320,7 @@ public class NativeGitProvider extends GitDataProvider {
 
   private String runGitCommand(File directory, long nativeGitTimeoutInMs, String gitCommand) throws NativeCommandException {
     final String env = System.getenv("GIT_PATH");
-    final String exec = (env == null) ? "git" : env;
+    String exec = env == null ? "git" : env;
     final String command = String.format("%s %s", exec, gitCommand);
 
     try {
@@ -358,11 +355,12 @@ public class NativeGitProvider extends GitDataProvider {
     private final String stdout;
     private final String stderr;
 
-    public NativeCommandException(int exitCode,
-                                  String command,
-                                  File directory,
-                                  String stdout,
-                                  String stderr) {
+    public NativeCommandException(
+            int exitCode,
+            String command,
+            File directory,
+            String stdout,
+            String stderr) {
       this.exitCode = exitCode;
       this.command = command;
       this.directory = directory;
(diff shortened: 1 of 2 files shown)


andreas-eternach:
I fixed the conflicts locally, there came up two more questions when doing that:
* the timeout did not fix the hanging process, because even after the timeout the streams where completely flushed (depending on the stream which caused the overflow, because the implementation processes only one of them, not in parallel)
* the new implementation is still not completely safe, there is a risk of an OutOfMemoryException in case the changes are really huge (larger than the heap).

TheSnoozer:
Thanks for creating a PR for that (greatly appreciated)!
I#ll try to review those changes on the weekend (sorry for the delay).

TheSnoozer:
Hi,
thanks again for the merge request and sorry for the delay on the review.
It honestly took me some time to review the changes, since it seems that you just moved some methods around without any actual changes.
I find it hard to read diffs where most of the stuff was just moved to some other place without any actual change (sorry, but its just super hard). For better readability I condensed the current changes down to [this](https://github.com/ktoso/maven-git-commit-id-plugin/files/2570318/patch-396.txt) that IMHO creates a better visibility what has changed exactly.

The BigDiff Tests passes, but that passed manually on my side before any changes. From my perspective the only thing I currently would ask for is to condense down the current MR to either the version I created (hopefully i didn't miss anything), or achieve the same in any other shape or form. Regardless on how you want to proceed I think it should also be verified that the bug is gone (that's the entire goal here). Besides that the changes seem reasonable and thus i would be to happy to merge the condensed form (where one can see in the diff what changed).

andreas-eternach:
Indeed the methods have been rearranged by my IDE automatically, this is a setting i am using for other projects, which is somehow not suitable for multi-peer-projects. I am sorry for that. 

About the BigDiffTest: its also not reliable reproducing the issue on my machine. In 1 of 4 cases it passes with the old implementation. Not sure why, i assume that the JDK dynamically allocates pipe-sizes. 

I try to look into your condensed patch later on this week. BR

TheSnoozer:
The auto-formatting changes happen for me all the time too....so nothing to be sorry about it happens :-)

Interesting, let me know how you would like to proceed and in particular if the final patch (whatever we conclude is the right thing) works as expected.

andreas-eternach:
Hello again, i fixed the test case, you could probably not repdroduce because it used jgit instead of native git. This change probably happened when i merged HEAD into the PR.

I now reverted the PR via a force-push, actually using your cleaned patch, only additional with the BigDiffTest, which i enabled and which should pass now.

src/main/java/pl/project13/maven/git/NativeGitProvider.java
```diff
@@ -443,19 +443,29 @@ public boolean runEmpty(File directory, long nativeGitTimeoutInMs, String comman
         // this only works on UNIX like system not on Windows
         // ProcessBuilder builder = new ProcessBuilder(Arrays.asList("/bin/sh", "-c", command));
         // so use the same protocol as used in the run() method
-        ProcessBuilder builder = new ProcessBuilder(command.split("\\s"));
+        final ProcessBuilder builder = new ProcessBuilder(command.split("\\s"));
         final Process proc = builder.directory(directory).start();
-        final InputStream is = proc.getInputStream();
-        final InputStream err = proc.getErrorStream();
-        BufferedReader reader = new BufferedReader(new InputStreamReader(is));
+        final SubProcessOutputGobbler subProcessOutputGobbler = new SubProcessOutputGobbler(proc.getInputStream());
```
TheSnoozer: Mhhh....since this code is now essentially somewhat is duplicated I'm starting to wonder what we actually need to fix the problem and also somewhat get rid of the code duplication here. IMHO keeping in mind that one always needs to adjust two part of the code is just a pain. Especially now since the complexity will increase.

I think in the end of the day to fix the problem we just need to read the `proc.getInputStream()` and `proc.getErrorStream()` in a separate thread (really like what is being done here inside the MR that launches a thread for each of them).

After thinking back and forth I again started to follow a different approach (sorry)...Let me elaborate...I currently see it as problem where we have the `run` function that would like to obtain all output and the `runEmpty` function that just wants to know if there is one line of output. Besides that both functions should pretty much do the same stuff with building a process and now building a thread structure to read `proc.getInputStream()` and `proc.getErrorStream()`.

In general I now tried to factor out the entire logic of building the thread structure and essentially ended up with creating a function that accepts a string (e.g. the content we read from either `proc.getInputStream()` or `proc.getErrorStream()`) and that function can return a boolean to indicate if it wants to continue reading the stream, or doesn't. The signature boils down to ``Function<String, Boolean>`` and converts the both read functions into the following:
```java
@Override
    public String run(File directory, long nativeGitTimeoutInMs, String command) throws IOException {
      String output = "";
      try {
        final StringBuilder commandResult = new StringBuilder();

        final Function<String, Boolean> stdoutConsumer = line -> {
          if (line != null) {
            commandResult.append(line).append("\n");
          }
          // return true to indicate we want to read more content
          return true;
        };
        runProcess(directory, nativeGitTimeoutInMs, command, stdoutConsumer);

        output = commandResult.toString();
      } catch (final InterruptedException ex) {
        throw new IOException(ex);
      }
      return output;
    }

    @Override
    public boolean runEmpty(File directory, long nativeGitTimeoutInMs, String command) throws IOException {
      final AtomicBoolean empty = new AtomicBoolean(true);

      try {
        final Function<String, Boolean> stdoutConsumer = line -> {
          if (line != null) {
            empty.set(false);
          }
          // return false to indicate we don't need to read more content
          return false;
        };
        runProcess(directory, nativeGitTimeoutInMs, command, stdoutConsumer);
      } catch (final InterruptedException ex) {
        throw new IOException(ex);
      }
      return empty.get(); // was non-empty
    }
```

All the we "build a thread structure" and "process error handling" is residing within `runProcess` and thus providing the same code to both functions.
See [here for for full reference](https://github.com/git-commit-id/maven-git-commit-id-plugin/files/2602607/396.txt).

WDYT? Does this look reasonable? Would that fix the problem? (Again, on my side everything seems to be working...)

andreas-eternach: As far as a i can see your modifications seem reasonable. So feel free to continue and use my PR as base line for your refactorings, which really look nice 👍

TheSnoozer: Thanks for your feedback and especially to your contribution on this project!
Will do that :-)

Issue #389, opened 2018-10-02
<replacementProperty> fails to work with an empty <value> tag


toolforger:
### Describe the bug
Tried to eliminate the `-SNAPSHOT` suffix from `git.build.version`, using an empty `<value>` tag.
The plugin would not apply the rule in that case (which does not make any sense).

### Steps to Reproduce; Expected and Observed Behaviour

Given a tag of "10.0-SNAPSHOT", this produces `git.build.version.nosnapshot=10.1` as intended:

```xml
           <plugin>
                <groupId>pl.project13.maven</groupId>
                <artifactId>git-commit-id-plugin</artifactId>
                <version>2.2.5</version>
                <executions>
                    <execution>
                        <goals>
                            <goal>revision</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <replacementProperties>
                        <replacementProperty>
                            <property>git.build.version</property>
                            <token>(.)-SNAPSHOT$</token> <!-- workaround: detect any character -->
                            <value>$1</value> <!-- workaround: have a nonempty <value> tag -->
                            <propertyOutputSuffix>nosnapshot</propertyOutputSuffix>
                        </replacementProperty>
                    </replacementProperties>
                </configuration>
            </plugin>
```

If I use

```xml
                            <token>-SNAPSHOT$</token>
                            <value></value>
```

or

```xml
                            <token>-SNAPSHOT$</token>
                            <value/>
```

I get `git.build.version.nosnapshot=10.1-SNAPSHOT`, i.e. the replacement is not happening.

No error messages or stack traces.

### Additional context
For reproducibility please provide the following:

JDK Version: 1.8.0_152
Maven Version: 3.3.9
OS Info: Arch: amd64 Family: windows Name: windows 7 Version: 6.1
[...]

PR #400, opened 2018-12-07
#389 allow replacementProperty value to be empty

jonevn:
### Context
Allow for value to be empty in replacement property.

https://github.com/git-commit-id/maven-git-commit-id-plugin/issues/389

### Contributor Checklist

### DIFF
--- a/src/main/java/pl/project13/maven/git/PropertiesReplacer.java
+++ b/src/main/java/pl/project13/maven/git/PropertiesReplacer.java
@@ -95,19 +95,19 @@ public class PropertiesReplacer {
   }
 
   private String replaceRegex(String content, String token, String value) {
-    if ((token == null) || (value == null)) {
-      log.error("found replacementProperty without required token or value.");
+    if (token == null) {
+      log.error("found replacementProperty without required token.");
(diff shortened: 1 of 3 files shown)


TheSnoozer:
Looks good to me.
Would you also tell maven that this property is not required? (see https://github.com/git-commit-id/maven-git-commit-id-plugin/blob/master/src/main/java/pl/project13/maven/git/ReplacementProperty.java#L62)

This essentially suffers from
* https://issues.apache.org/jira/browse/MNG-6434
* https://issues.apache.org/jira/browse/MNG-5725

TheSnoozer:
On Congratulation for the 400th issue/merge request :-)

jonevn:
I will look into the maven property required thing. Do we want it to be possible to completely skip the value-tag?

TheSnoozer:
Hi,
thanks again for the changes and sorry for the delay here. Looks good to me. I guess the default value for some nested stuff might not work properly (for the same reason the required is not working -- so the value will end up being null instead of the default). However I guess in that case the additional null check that is present should 'fix' this.

Merging.

PR #424, opened 2019-07-12
Introduce "offline" mode

hgschmie:
Version 3.0.0 started to look at "ahead" and "behind" branches. To do
so, it executes a Fetch command to the repo. This is unfortunate
because it introduces quite a bit of a delay when building, especially
when the machine is offline (e.g. on a plane). Also, it sets off
various alerts because the previous version did not do any remote
access.

This PR introduces an "offline" mode (disabled by default) that allows
turning off all commands that are trying to reach out to
remotes. Currently, that is only the fetch command which is executed
when trying to determine "ahead" and "behind". Any future operation
that is accessing remote repos should also check this flag.

### Contributor Checklist

### DIFF
--- a/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
+++ b/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
@@ -34,6 +34,7 @@ import org.apache.maven.plugins.annotations.LifecyclePhase;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
+import org.apache.maven.settings.Settings;
 import org.sonatype.plexus.build.incremental.BuildContext;
 
 import pl.project13.maven.git.build.BuildServerDataProvider;
@@ -69,6 +70,12 @@ public class GitCommitIdMojo extends AbstractMojo {
    */
   @Parameter(property = "session", required = true, readonly = true)
   MavenSession session;
+
+  /**
+   * The Maven settings.
+   */
+  @Parameter(property = "settings", required = true, readonly = true)
+  Settings settings;
 
   /**
    * <p>Set this to {@code 'true'} to inject git properties into all reactor projects, not just the current one.</p>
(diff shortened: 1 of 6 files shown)


TheSnoozer:
Hi, thanks for your contribution!

I'd agree that this `fetch` is certainly not the taste of everyone and thus was highlighted in the release notes.

Does the exclusion of the ahead and behind properties still trigger a fetch?
E.g.
```
<excludeProperties><excludeProperty>^git.local.branch.*$</excludeProperty></excludeProperties>
```
?

hgschmie:
Looking through the code, it seems that the `maybePut` in `GitDataProvider` would not trigger the memoizer for fetching the remote branches if the ahead and behind branches are not accessed. Now, given that this uses the slightly less than obvious syntax of excluding `git.local.branch.*$` (so it excludes the local branches to not trigger accessing remote branches???), I still feel it would be justified to have an actual switch to the configuration of the plugin.

This switch would also allow to be triggered in a profile to turn remote access on and off (I use that in connection with properties in the basepom project all the time).

I also updated the PR to respect the maven "-o" offline switch to put maven in offline mode.

TheSnoozer:
Hi, thanks for your input. 
AFAIK with that exclusion the plugin should also not trigger a pull and thus was mainly raising the question if this plugin really needs yet another switch (considering the many's that are already there). 
However I think you really point out the problem with that slightly less obvious syntax and just referencing to an older release note (https://github.com/git-commit-id/maven-git-commit-id-plugin/releases/tag/v3.0.0) where I mentioned this could easily be overlooked. Considering the update to respect the  maven "-o" offline makes it even more obvious that there is a need for such new setting.

src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
```diff
@@ -70,6 +71,12 @@
   @Parameter(property = "session", required = true, readonly = true)
   MavenSession session;
 
+  /**
+   * The Maven settings.
+   */
+  @Parameter(property = "settings", required = true, readonly = true)
+  Settings settings;
```
TheSnoozer: This new settings sadly makes the [tests fail](https://travis-ci.org/git-commit-id/maven-git-commit-id-plugin/builds/559300144).

Inside the [GitIntegrationTest](https://github.com/git-commit-id/maven-git-commit-id-plugin/blob/fc4f5e8e059d40301a8096f93abd2c687cae7b76/src/test/java/pl/project13/maven/git/GitIntegrationTest.java#L118) you also would need to mock them:
```java
  private static Settings mockSettings() {
    Settings settings = mock(Settings.class);
    when(settings.isOffline()).thenReturn(false);
    return settings;
  }
```

hgschmie:
Hey, sorry for breaking the tests. I added the change to the code; it now builds on my laptop (and also on the CI).

TheSnoozer:
Perfect! Thank you for your contribution and your patience to address my comments ;-)

I'm going ahead and merge. This feature will become available in the next version.

Issue #354, opened 2018-01-23
Use of java.net.InetAddress.getLocalHost().getHostName() can be very slow


vorburger:
### Expected Behavior
maven-git-commit-id-plugin should work well offline and be fast! :smile: 

### Current Behavior
The use of `java.net.InetAddress.getLocalHost().getHostName()` in `GitCommitIdMojo.loadBuildHostData` can be slow on certain networks (e.g. bad WiFi); as seen e.g. in https://github.com/vorburger/opendaylight-eclipse-setup/issues/4.  Various online posts document why that is so - it does a reverse DNS look up.

There are libraries with native equivalents; I don't know if you would consider using one of those?

Or perhaps there could simply be a property to skip `loadBuildHostData` ?

TheSnoozer:
Hi, thanks for your report and the detailed investigation :
)
There are certain things that might take a long time with this plugin and might not be needed by all consumers.
Introducing another switch would be somewhat painfull and would be potentially needed for every property.
Plus in all honesty I think this plugin has way to many switched and tweeks already and as a result I would try to avoid to introduce another configuration option.

However I fully agree that there should be something to avoid getting stuck in any method when the data is not needed. Would https://github.com/ktoso/maven-git-commit-id-plugin/issues/185 be something that would help to solve this issue in your case?

vorburger:
@TheSnoozer yeah a way to just exclude the `build.host` to save the `loadBuildHostData` which does that `java.net.InetAddress.getLocalHost().getHostName()` would address this, so if you want to close this as a dupe of #185 that's perfectly fine with me.

TheSnoozer:
Alright, I think its time to get this selective running shipped! 
There are enough reasons (speedup and failures in certain extraction methods) to justify its benefits.

I'll close this as duplicate and will try to find some time to fix this (and other minor bugs) over the next days.

Thanks again for reporting this!

Issue #185, opened 2015-04-10
Instead of determine all properties and then exclude properties why not choose selective running (native git)


TheSnoozer:
Hi,
currently the plugin has the option to exclude some properties or not. This is really usefull and I use this on a daily basis (yeah customers don't want to have some properties in some sensitive files).
However, for the native git the plugin currently determines all properties with calls to the git executable.
Each call coast a few milli seconds build time. So what the native git implementation currently does it will call the git executable, determine the result and when this is done the properties get's filtered out if they should be excluded.
Why not just skip the call to the binary, when the property isn't wanted anyway?
For each property this would speed up the runtime for the native git implementation.

As far as I know changing the behavoir for the jgit-implementation wouldn't speed up the build process. This is due to the fact that the jgit implementation just gathers all information that it could get and the maven-git-commit-id-plugin just asks for them afterwards (as far as I know there is no second call to the jgit-implementation).

Just to get the Idea why I'm suggesting this:
I'm using the native git and some similar config to this:

```
<excludeProperties>
                        <excludeProperty>git.branch</excludeProperty>
                        <excludeProperty>git.build.*</excludeProperty>
                        <excludeProperty>git.commit.user.*</excludeProperty>
                        <excludeProperty>git.commit.message.*</excludeProperty>
                        <excludeProperty>git.commit.time</excludeProperty>
                        <excludeProperty>git.commit.id.describe-short</excludeProperty>
                        <excludeProperty>git.remote.origin.url</excludeProperty>
</excludeProperties>
```

this means exclude everything and just keep git.commit.id and git.commit.id.abbrev....
so in my case I would speed up the build process by probability 1s :D

goranmagdic:
Hi, 

I find plugin very slow in my case and I only need commit.id information, so I assume it would be much faster if it did not parse the whole tag history. is there maybe something already that I am missing?

TheSnoozer:
Can you elaborate further on what "slow" means for you?
Can you paste your full configuration here?
Are you doing a shallow clone or full clone of the git repository?
Have you tried the native git implementation?

goranmagdic:
Just counting while maven is stuck on plugin I got to 9 sec. That is for one module. 
I can paste plugin configuration, not the whole project: 
`<plugin>
			<groupId>pl.project13.maven</groupId>
			<artifactId>git-commit-id-plugin</artifactId>
			<version>2.2.1</version>
			<executions>
				<execution>
[...]

PR #428, opened 2019-08-10
Add filtering capabilities for build server data.

ryantse:
### Context
The changes introduced in #185 do not cover the issue reported in #354 where a call to `getHostName()` may take a long time when getting the property `git.build.host`. Likewise, in certain (rare) cases, `TimeZone.getTimeZone()` may also have an impact upon performance when retrieving `git.build.time`. 

Depending on the number of projects, this may have a large effect on compile times as there may be numerous calls to the affected functions. A user should be able to filter these properties from being fetched so that build times can improve if this becomes an issue. This also further aligns with the changes in #185 to be more consistent with not fetching unused properties.

### Contributor Checklist

### DIFF
--- a/src/main/java/pl/project13/maven/git/build/BuildServerDataProvider.java
+++ b/src/main/java/pl/project13/maven/git/build/BuildServerDataProvider.java
@@ -19,6 +19,7 @@ package pl.project13.maven.git.build;
 
 import org.apache.maven.project.MavenProject;
 import pl.project13.maven.git.GitCommitPropertyConstant;
+import pl.project13.maven.git.PropertiesFilterer;
 import pl.project13.maven.git.log.LoggerBridge;
 import pl.project13.maven.git.util.PropertyManager;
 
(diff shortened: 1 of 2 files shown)


src/main/java/pl/project13/maven/git/build/BuildServerDataProvider.java
```diff
@@ -152,7 +169,7 @@ protected void maybePut(@Nonnull Properties properties, @Nonnull String key, Sup
     if (properties.contains(keyWithPrefix)) {
       String propertyValue = properties.getProperty(keyWithPrefix);
       log.info("Using cached {} with value {}", keyWithPrefix, propertyValue);
-    } else {
+    } else if (PropertiesFilterer.isIncluded(keyWithPrefix, includeOnlyProperties, excludeProperties)) {
```
TheSnoozer: Good catch!

TheSnoozer:
Thank you for your contribution!

Indeed this seems and sounds reasonable. I personally honestly don't know in which weird cases `TimeZone.getTimeZone()` may also have an impact upon performance. I guess this is not part of the question/problem and out of scope for this MR.


Changes look good for me. I'm going ahead and merge.
Thank you again for your contribution!

PR #429, opened 2019-08-10
Fix property caching when using injectAllReactorProjects.

ryantse:
### Context
When checking for whether a key is within the `Properties` object, using `.contains` is equivalent of `.containsValue`, while the correct invocation is `.containsKey`. This also fixes the logging when saving the `Properties` object to the context.

### Contributor Checklist


TheSnoozer:
Hello,
thank you again for your contribution!
This looks very reasonable and looks like a small bug that makes the entire caching not work (sad times) :/

I'm going ahead and merge ;-)

Thanks!

PR #454, opened 2019-11-29
Add support for GitHub Actions, some refactoring

andrebrait:
### Context
This adds support for GitHub Actions and refactors a few things.

### Contributor Checklist
- [:x:] Added relevant integration or unit tests to verify the changes
- [:heavy_check_mark:] Update the Readme or any other documentation (including relevant Javadoc)
- [:heavy_check_mark:] Ensured that tests pass locally: `mvn clean package`
- [:heavy_check_mark:] Ensured that the code meets the current `checkstyle` coding style definition: `mvn clean verify -Pcheckstyle -Dmaven.test.skip=true -B`

### DIFF
--- a/core/src/main/java/pl/project13/core/cibuild/GitHubBuildServerData.java
+++ b/core/src/main/java/pl/project13/core/cibuild/GitHubBuildServerData.java
@@ -1,0 +1,65 @@
+/*
+ * This file is part of git-commit-id-plugin by Konrad 'ktoso' Malawski <konrad.malawski@java.pl>
+ *
+ * git-commit-id-plugin is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * git-commit-id-plugin is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with git-commit-id-plugin.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package pl.project13.core.cibuild;
+
+import pl.project13.core.log.LoggerBridge;
+
+import javax.annotation.Nonnull;
+import java.util.Map;
+import java.util.Properties;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+
+public class GitHubBuildServerData extends BuildServerDataProvider {
+
+    private static final String BRANCH_REF_PREFIX = "refs/heads/";
+    private static final String PULL_REQUEST_REF_PREFIX = "refs/pull/";
+
+    GitHubBuildServerData(LoggerBridge log, @Nonnull Map<String, String> env) {
+    super(log,env);
+  }
+
+  /**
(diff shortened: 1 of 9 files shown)


core/src/main/java/pl/project13/core/cibuild/GitHubBuildServerData.java
```diff
@@ -0,0 +1,65 @@
...
+  @Override
+  public String getBuildBranch() {
+    String environmentBasedBranchKey = "GITHUB_REF";
+    String environmentBasedBranch = env.get(environmentBasedBranchKey);
+    if (!isNullOrEmpty(environmentBasedBranch)) {
+      if (environmentBasedBranch.startsWith(BRANCH_REF_PREFIX)) {
+        environmentBasedBranch = environmentBasedBranch.substring(BRANCH_REF_PREFIX.length());
+      } else if (environmentBasedBranch.startsWith(PULL_REQUEST_REF_PREFIX)) {
+        environmentBasedBranchKey = "GITHUB_HEAD_REF";
+        environmentBasedBranch = env.get(environmentBasedBranchKey);
```
TheSnoozer: Interesting, based on the doc this appears be the branch-name and thus would not be prefixed with `refs/heads/` - IMHO this makes the similar sounding property somewhat inconsistent with the `GITHUB_REF` that is prefixed `refs/heads/`.

Anyways I guess its not a mission critical property...

TheSnoozer:
Hello,
thank you for your contribution!

Looks good for me, so I'm just going ahead and merge :-)

core/src/main/java/pl/project13/core/cibuild/GitHubBuildServerData.java
andrebrait: `GITHUB_HEAD_REF` is not prefixed in any way. Maybe my code was not very readable since I reused the variable name, I guess.

Here I'm checking if `GITHUB_REF` is prefixed with `refs/heads/`, in which case I use its value itself as the name of the branch (minus the `refs/heads/` part). 

Otherwise, if it's a pull request (i.e. `GITHUB_REF` is prefixed with `refs/pull/`) then I get the name from `GITHUB_HEAD_REF` *unmodified*.

You can see some tests here:

[Commit to `master` branch](https://github.com/andrebrait/actions_test/commit/0c5496f51cbff741c0436f41802bd64550132139/checks?check_suite_id=334590862):
```properties
GITHUB_REF=refs/heads/master
```

[Pull request from `andrebrait-patch-1` to `master`](https://github.com/andrebrait/actions_test/commit/0cb5b11732d0d0bf9d98758b437f7fbf102b7a4f/checks?check_suite_id=334578295):
```properties
GITHUB_REF=refs/pull/1/merge
GITHUB_HEAD_REF=andrebrait-patch-1
```

andrebrait: I created #455 to make it more explicit and also fix a possible issue when building from tags.

TheSnoozer: Thanks for the follow up :-)
To be fair I found your code readable enough, my comment was more regarding the naming of the variables from the github side....both are named `ref` creating the impression that both should be formatted similar...anyhow let me look at your MR :-)

andrebrait: Ah, yes. Indeed. The whole set of environment variables that they use is a bit weird, IMHO. Getting the branch name should not require one to parse it, for example.

TheSnoozer: Yeah agree...as far as I can recall this plugin also needs to remove the `refs/heads` from some output of jgit so from that perspective its nothing out of the ordinary...and github actions are quite new so i would imagine they might introduce or change it after getting some feedback

PR #455, opened 2019-11-29
Refactor for better readability, fix returning branch name based on tags

andrebrait:
I think this makes the code cleaner, and fixes a case where the returned branch name could be `refs/tags/<tagname>`.


TheSnoozer:
Thanks for the follow up :-)

LGTM :-)

PR #464, opened 2020-01-05
[tests] Migrate off fest-assert to assertj

hazendaz:
fest-assert 2 once was assertj.  The authors differed in direction and the framework split.  fest-assert itself is long dead.  Version 2 never made release as a result and the migration as it were is incredibly easy as seen here and at least in one test set makes more logical sense.

### DIFF
--- a/maven/pom.xml
+++ b/maven/pom.xml
@@ -35,7 +35,7 @@
     <junit.version>4.13.1</junit.version>
     <mockito.version>3.1.0</mockito.version>
 
-    <fest-assert.version>1.4</fest-assert.version>
+    <assertj.version>3.14.0</assertj.version>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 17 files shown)


hazendaz:
I'm looking into the failures now.  I had another dozen or so commits past this on my local and it built fine.  I'm tracking down the issue now but at least looking at those tests, it seems clear the item it wants isn't actually there.  It's also extremely slow to build (10 minutes) without my other changes so I kind of think something else is at play.  I'll get back to this shortly.

hazendaz:
found issue later in my code but actually a direct problem in current master here.  junit 5 is not just an api.  It requires the engine or it's simply not even used.  I had later updated surefire which hard requires to konw what its doing.  As such it will skip all tests if either jupiter or vintage are not defined.  As this code base is NOT actually on junit 5 at all, it needed to define the vintage engine not the jupiter api.  Anyway, I'm fixing that too and will add to this PR while I look at the 2 specific tests that are failing.

hazendaz:
fixed test issue (used wrong check).

hazendaz:
Rebased

TheSnoozer:
I somewhat fully lost track of this open MR (sorry for that).

Changes look good for me! Thanks again for this effort!

hazendaz:
No worries :)  I lost track of it too.  Thanks.

PR #486, opened 2020-05-13
Simplify GitHub workflow, cache local Maven repository

Marcono1234:
### Context
Changes the GitHub workflow:
- Remove Java version matrix where it makes sense to only use one version, e.g. deploying snapshot
- Use consistent `mvn` command structure:
  - `-B` as first flag
  - `--file pom.xml` is redundant
- Remove "Package with Maven" step from `integration-test` because `clean install` should cover that already
- Use cache to not require GitHub to download Maven dependencies every time

Note that concurrent / parallel usage of the same cache might cause issues (see https://github.com/actions/cache/issues/189), however without it, a cache per job, per Java version / Maven version would be needed, for which the total cache size might become pretty large at some point.

You can see a successful run in my fork: https://github.com/Marcono1234/git-commit-id-maven-plugin/actions/runs/103612944

### DIFF
--- a/.github/workflows/default-tests.yml
+++ b/.github/workflows/default-tests.yml
@@ -4,23 +4,26 @@ on: [push, pull_request]
 
 jobs:
   checkstyle:
-    name: Run checkstyle with java ${{ matrix.java_version }}
+    name: Run checkstyle
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


TheSnoozer:
Hello,
looks reasonable for me. Thank you for the simplifications and your contribution!

PR #492, opened 2020-06-10
Remove unused Java version matrix from GitHub workflow

Marcono1234:
### Context
The Java version matrix is not used anymore because the Java version is specified explicitly. I forgot to remove that in #486.


TheSnoozer:
Thanks for the extra removal! LGTM

PR #493, opened 2020-06-12
Remove dependency on Google Guava

joschi:
### Context

While Google Guava provides a lot of useful functionality, it is a rather heavy-weight dependency and was notorious for breaking backward-compatibility in the past.

There are only very few features from Google Guava used in `git-commit-id-maven-plugin` which could either be replaced by native Java 8 features (`Files.createParentDirs()`) or removed altogether (`@VisibleForTesting`).

### Contributor Checklist

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -21,10 +21,6 @@
     </properties>
 
     <dependencies>
-        <dependency>
-            <groupId>com.google.guava</groupId>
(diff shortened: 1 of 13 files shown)


TheSnoozer:
Thank you for your contribution! Looks good for me! Thanks!

PR #496, opened 2020-06-12
Add support for AWS CodeBuild

joschi:
### Context

Add support for fetching build information on [AWS CodeBuild](https://aws.amazon.com/codebuild/).

The available environment variables in the AWS CodeBuild build environment are listed at [Environment variables in build environments](https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html).

### Contributor Checklist
- [x] Update the Readme or any other documentation (including relevant Javadoc)

### DIFF
--- a/core/src/main/java/pl/project13/core/cibuild/AwsCodeBuildBuildServerData.java
+++ b/core/src/main/java/pl/project13/core/cibuild/AwsCodeBuildBuildServerData.java
@@ -1,0 +1,55 @@
+/*
+ * This file is part of git-commit-id-plugin by Konrad 'ktoso' Malawski <konrad.malawski@java.pl>
+ *
+ * git-commit-id-plugin is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * git-commit-id-plugin is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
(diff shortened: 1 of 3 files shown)


core/src/main/java/pl/project13/core/cibuild/AwsCodeBuildBuildServerData.java
```diff
@@ -0,0 +1,55 @@
...
+  public static boolean isActiveServer(Map<String, String> env) {
+    return env.containsKey("CODEBUILD_BUILD_ARN");
+  }
+
+  @Override
+  void loadBuildNumber(@Nonnull Properties properties) {
+    String buildNumber = env.getOrDefault("CODEBUILD_BUILD_NUMBER", "");
+    put(properties, GitCommitPropertyConstant.BUILD_NUMBER, buildNumber);
+
+    String buildArn = env.get("CODEBUILD_BUILD_ARN");
```
TheSnoozer, line 45: Without having looked into this further and just reading the properties on https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-env-vars.html the `CODEBUILD_BUILD_ARN` contains `region-ID:account-ID`. While the region might not contain sensitve information (e.g. could simply be `us-east-1`) I'm not so sure about the `account-ID`. For me it seems that the `CODEBUILD_BUILD_ID` provides the 'uniqueness' desired here and thus I'd suggest to use this instead.

```suggestion
    String buildArn = env.get("CODEBUILD_BUILD_ID");
```

joschi: @TheSnoozer Sounds good to me.

Strictly speaking the build ID is not unique because it could be reused across regions and AWS accounts but in reality this probably shouldn't matter. 👍

TheSnoozer:
Thanks again for your contribution! Looks good now, I'm going ahead and merge.

PR #501, opened 2020-06-27
replace jackson-databind with lightweight javax.json

TheSnoozer:
### Contributor Checklist

Perhaps notifying @joschi which might create conflicts with the effort started in https://github.com/etiennestuder/java-ordered-properties/issues/4


joschi:
@TheSnoozer Is there any inherent advantage in using `javax.json` over Jackson?
The API looks very similar to [`jackson-core`](https://github.com/FasterXML/jackson-core) (on which `jackson-databind` builds) and it's an external dependency either way.

TheSnoozer:
Mhh I don't think there is any advantage in using `javax.json` over `jackson-core`. Almost all APIs I have seen offer what the plugin needs and perhaps the `javax.json` variant is more common. I'm still somewhat disappointed that java8 doesn't seem to offer anything out of the box. 

Perhaps as background for the replacement of `jackson-databind`: I'm just annoyed by the amount of CVE's. While this project might not be effected by any of them, this project still depends on `jackson-databind` and will get flagged with the CVE when they pop up. IMHO this requires explaining why this project is not effected when someone actually looks at the report....so in short I think I just want to get rid of `jackson-databind`. For the replacement it really can be any API that does what the plugin needs (generate JSON from a properties object/read JSON to a properties object).

joschi:
> Perhaps as background for the replacement of `jackson-databind`: I'm just annoyed by the amount of CVE's.

FWIW, pretty much all of them are issued for Jackson 2.9 and the "offending" feature has been fixed in Jackson 2.10 and 2.11: https://medium.com/@cowtowncoder/jackson-2-10-safe-default-typing-2d018f0ce2ba

If you're looking for a light-weight alternative to `jackson-databind` without going as low-level as `jackson-core` or `javax.json`, there's [`jackson-jr`](https://github.com/FasterXML/jackson-jr) to fill that gap.

> Almost all APIs I have seen offer what the plugin needs and perhaps the `javax.json` variant is more common.

Whatever the stats are worth, on https://mvnrepository.com/ it looks like ~900 usages of [`javax.json`](https://mvnrepository.com/artifact/org.glassfish/javax.json) and ~7000 ([`jackson-core`](https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-core)/~15000 ([`jackson-databind`](https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-databind)) usages of Jackson.

This all being said, I don't want to talk you out of that change but merely provide a second perspective. 😉

PR #502, opened 2020-06-29
Enable reproducible builds by suppressing date comment in properties

### LINKED ISSUE #472: Prevent header timestamp in git.properties to allow for reproducible builds
New feature: Prevent header timestamp in git.properties to allow for reproducible builds.

The git.properties file output includes the current time as a timestamp in a commented header.  This prevents reproducible builds because each build will include a different timestamp.

A feature fix would prevent this timestamp from being written as a comment header.

Note1: For reproducible builds the $git.build.time parameter also needs to be excluded
Note2: Using json format is a workaround for this issue

For reference:

https://maven.apache.org/guides/mini/guide-reproducible-builds.html

p.s.  Thanks for the awesome plugin and especially the detailed configuration documentation!

joschi:
### Context

The native `java.util.Properties` class in Java will add a comment in each properties file with the timestamp when the file was generated, which prevents projects implementing [reproducible builds](https://reproducible-builds.org/).

This change set replaces the usage of `java.util.Properties` and the custom order-preserving wrapper around it with the [`java-ordered-properties`](https://github.com/etiennestuder/java-ordered-properties) project, which prevents writing the timestamp comment and still retains a reproducible order in the properties files.

Fixes #472

### Contributor Checklist


TheSnoozer:
Hi, sorry for the delay (again). Thanks for taking care of this and your contribution! Also a great shout-out for your contribution to https://github.com/etiennestuder/java-ordered-properties/issues/4!

Changes look good to me. I'm going ahead and merge :-)

PR #511, opened 2020-07-26
Build system information should be optionally set

### LINKED ISSUE #469: Not thread safe when using injectAllReactorProjects.
### Describe the bug
When using the option `injectAllReactorProjects` the plugin is not thread safe when accessing the shared Property object. This causes a concurrent modification exception when doing multi-threaded builds.

### Steps to Reproduce
Invoke `mvn install` with a high number of projects and enable multi-threaded builds.

```xml
<plugin>
  <executions>
    <execution>
      <id>for-jars</id>
      <inherited>true</inherited>
      <goals>
        <goal>revision</goal>
      </goals>
      <configuration>
        <generateGitPropertiesFilename>target/classes/git.properties</generateGitPropertiesFilename>
      </configuration>
    </execution>
    <execution>
      <id>for-source-tarball</id>
      <goals>
        <goal>revision</goal>
      </goals>
      <inherited>false</inherited>
      <configuration>
        <generateGitPropertiesFilename>./git.properties</generateGitPropertiesFilename>
      </configuration>
    </execution>
  </executions>
  <configuration>
    <dateFormat>yyyy-MM-dd'T'HH:mm:ssZ</dateFormat>
    <verbose>true</verbose>
    <skipPoms>false</skipPoms>
    <generateGitPropertiesFile>true</generateGitPropertiesFile>
    <failOnNoGitDirectory>false</failOnNoGitDirectory>
    <useNativeGit>true</useNativeGit>
    <injectAllReactorProjects>true</injectAllReactorProjects>
    <includeOnlyProperties>
      <includeOnlyProperty>git.build.time</includeOnlyProperty>
      <includeOnlyProperty>git.build.user.email</includeOnlyProperty>
      <includeOnlyProperty>git.commit.id</includeOnlyProperty>
      <includeOnlyProperty>git.commit.time</includeOnlyProperty>
    </includeOnlyProperties>
  </configuration>
</plugin>
```

- Is there a (public) project where this issue can be reproduced? No.
- Include any stack-traces or any error messages

### Expected behavior
Maven should not report a ConcurrentModificationException when running in parallel.

### Additional context
- git-commit-id-maven-plugin version 4.0.0
- Java version "1.8.0_221"
- Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
- Linux
- Bash
- mvn install

ryantse:
Changes the build system information to be conditionally set within the properties. In multi-threaded builds, without the optional put, a ConcurrentModificationException occurs as this information is written at the same time the object is being iterated over.

### Context
Fixes #469.

### Contributor Checklist

### DIFF
--- a/core/src/main/java/pl/project13/core/GitDataProvider.java
+++ b/core/src/main/java/pl/project13/core/GitDataProvider.java
@@ -245,7 +245,7 @@ public abstract class GitDataProvider implements GitProvider {
   protected void maybePut(@Nonnull Properties properties, String key, SupplierEx<String> value)
           throws GitCommitIdExecutionException {
     String keyWithPrefix = prefixDot + key;
-    if (properties.containsKey(keyWithPrefix)) {
+    if (properties.stringPropertyNames().contains(keyWithPrefix)) {
       String propertyValue = properties.getProperty(keyWithPrefix);
       log.info("Using cached {} with value {}", keyWithPrefix, propertyValue);
     } else if (PropertiesFilterer.isIncluded(keyWithPrefix, includeOnlyProperties, excludeProperties)) {
--- a/core/src/main/java/pl/project13/core/PropertiesFileGenerator.java
+++ b/core/src/main/java/pl/project13/core/PropertiesFileGenerator.java
@@ -75,8 +75,8 @@ public class PropertiesFileGenerator {
 
           final String buildTimeProperty = prefixDot + GitCommitPropertyConstant.BUILD_TIME;
 
-          propertiesCopy.remove(buildTimeProperty);
-          persistedProperties.remove(buildTimeProperty);
+          propertiesCopy.setProperty(buildTimeProperty, null);
+          persistedProperties.setProperty(buildTimeProperty, null);
 
           shouldGenerate = !propertiesCopy.equals(persistedProperties);
         } catch (CannotReadFileException ex) {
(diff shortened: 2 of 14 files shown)


joschi:
@ryantse As far as I see `maybePut` is also not thread-safe, so even with the change, we would see `ConcurrentModificationException` in some cases.

If it was possible at all, it would be great if we could get the build server data only once and then share it with the invocations of the plugin in the Maven sub-modules.

Otherwise making `BuildServerDataProvider` thread-safe would be an option too.

ryantse:
@joschi: Thanks for the quick feedback. You're right that the original patch didn't fully correct the thread-safety issue. The only time this really is a problem is when injectAllReactorProjects is turned on along with additional properties being configured in sub-modules.

I have pushed another patch that more comprehensively addresses the issue. This patch removes all usage of the underlying HashMap functions in preference for equivalents in the Properties class. As Properties is properly synchronized, this mitigates all instances where multiple modules (and thus multiple threads) share the same object.

ryantse:
@TheSnoozer: Could you take a look at this pull request as well?

ryantse:
@TheSnoozer: Can this be merged or do you have any feedback on this patch? Thanks in advance!

TheSnoozer:
Hi,
sorry for the delay here. Thanks for creating the Merge-Request with the needed changes. From what I can tell they look reasonable for me. Would this perhaps already mean we need to cut a 4.0.2 release (e.g. https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/507)? Perhaps could you confirm the changes fixed the issue locally? If so I'd just cut a release on the weekend.

ryantse:
@TheSnoozer: I was able to test the snapshot build of the plugin with our build system and it works well in my environment. It would be great if you could release a 4.0.2 with these changes.

TheSnoozer:
Thanks for confirming! Will do a release on the weekend.

TheSnoozer:
Now available as https://repo1.maven.org/maven2/pl/project13/maven/git-commit-id-plugin/4.0.2/. Thanks again for the investigation and pushing for the solution!

PR #512, opened 2020-08-03
Fix generated properties file NPE.

ryantse:
- Adds a test case to verify that the written properties file can be read again.

### Context
Accidentally introduced an NPE when reading an existing properties file and doing the comparison. Added a test case to ensure coverage of this case.

### Contributor Checklist


ryantse:
@TheSnoozer: Missed an edge case in my last pull request where setProperty cannot be passed a null. Instead, this change sets them to empty strings for the purposes of comparison. I've also added a test here to make sure that we test rereading the generated property file.

TheSnoozer:
Thanks for the follow up and adding a test-case :-)
Looks reasonable for me.

TheSnoozer:
Refs https://github.com/git-commit-id/git-commit-id-maven-plugin/pull/511

Issue #136, opened 2014-10-31
Enable per module versions


markokocic:
In multi module maven project that is stored in one Git repo it would be helpful to be able to get version number / build number based of the last git commit in the module itself, and not based on the last git commit of the whole git project.

That could even be a configuration option.

kdoteu:
We need this for our model repo. We have a multi module project with all necessary model modules we need across our projects. At this point we are using the lattest commit id of the repo. But to improve the quality an speed it would be a nice improvment to have the direct commit id of the module itself.

ktoso:
Would love to accept a contribution of such feature, you can start out by writing a test and take it from there though the Mojo class.
Currently not able time wise to work on it.

Thanks in advance!

PR #535, opened 2020-12-03
Add option for multi module projects to get version number / build nu…

nekrasov-iv:
…mber based of the last git commit in the module itself, and not based on the last git commit of the whole git project.

Enable per module versions #136 https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/136

### DIFF
--- a/core/src/main/java/pl/project13/core/GitDataProvider.java
+++ b/core/src/main/java/pl/project13/core/GitDataProvider.java
@@ -60,6 +60,8 @@ public abstract class GitDataProvider implements GitProvider {
 
   protected boolean offline;
 
+  protected String projectDirectory;
+
   public GitDataProvider(@Nonnull LoggerBridge log) {
     this.log = log;
   }
@@ -111,6 +113,11 @@ public abstract class GitDataProvider implements GitProvider {
 
   public GitDataProvider setOffline(boolean offline) {
     this.offline = offline;
+    return this;
+  }
+
+  public GitDataProvider setProjectDirectory(String projectDirectory) {
+    this.projectDirectory = projectDirectory;
     return this;
   }
 
--- a/core/src/main/java/pl/project13/core/JGitProvider.java
+++ b/core/src/main/java/pl/project13/core/JGitProvider.java
@@ -82,29 +82,52 @@ public class JGitProvider extends GitDataProvider {
   @Override
   public void prepareGitToExtractMoreDetailedRepoInformation() throws GitCommitIdExecutionException {
     try {
-      // more details parsed out bellow
-      Ref evaluateOnCommitReference = git.findRef(evaluateOnCommit);
-      ObjectId evaluateOnCommitResolvedObjectId = git.resolve(evaluateOnCommit);
-
-      if ((evaluateOnCommitReference == null) && (evaluateOnCommitResolvedObjectId == null)) {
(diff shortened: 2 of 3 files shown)


TheSnoozer:
Hello,
thank you very much for attempting to fix a very old issue on this project. Could you perhaps elaborate on your use-case for those changes?

To be honest I'm not even so sure if #136 is still an open issue. In specific I don't understand the difference to 'accepting multiple repositories' (outlined in https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/137#issuecomment-418144756). When seeing #136 as `run the plugin on every submodule` it raises question how the git prefix should be altered across projects (otherwise they would cancel each other out when we inject the properties into other projects) and it would create a conflict with the `runOnlyOnce`-idealogy (one would need to get a warning that combining the two doesn't make sense).

nekrasov-iv:
> 
> 
> Hello,
> thank you very much for attempting to fix a very old issue on this project. Could you perhaps elaborate on your use-case for those changes?

Hello,
thank you for taking the time to see my pull request.

In our case, we have a multi module project that looks like this:

module1
-pom.xml
module2
-pom.xml
...
moduleN
-pom.xml
pom.xml (the parent pom, also contains <modules> section with list of our modules)

We decided to include date and time of last change of module to the name of the module artefact. So the module jar (or war) file name should look like this module1-YYYYDDMM-HHMMSS.jar, where YYYYDDMM-HHMMSS is the date and time of the last commit in the module1 folder.

By default I get the date and time of the last commit in the repository, not in the module folder.

Maybe I missed something in the documentation and I can get this behavior using configuration?

TheSnoozer:
Just out of curiosity:
Where is your `.git`-folder or folders?

Does it look like:
```
+--- module1
|   +--- pom.xml
|    \--- `.git`
+--- module2
|   +--- pom.xml
|    \--- `.git`
....
+--- moduleN
|   +--- pom.xml
|    \--- `.git`
+---- pom.xml (the parent pom, also contains <modules> section with list of our modules)
\--- `.git` (root git directory)
```
and for each module you would like to get the date and time of last change related to the corresponding `.git` directory inside the module dir? So in practice you'd have N modules and thus N different timestamps? 

Or really just one extra `.git` directory 'under' module1 like:
```
+--- module1
|   +--- pom.xml
|    \--- `.git`
+--- module2
|    \--- pom.xml
....
+--- moduleN
|    \--- pom.xml
+---- pom.xml (the parent pom, also contains <modules> section with list of our modules)
\--- `.git` (root git directory)
```
?

If option 2 then I'd propose the settings as outlined in https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/137#issuecomment-418144756 where you would need to specify the `<dotGitDirectory>`.

nekrasov-iv:
We have only one `.git` directory, it is located at the root of the project:

```
+--- module1
|   +--- pom.xml
+--- module2
|   +--- pom.xml
....
+--- moduleN
|   +--- pom.xml
+---- pom.xml (the parent pom, also contains <modules> section with list of our modules)
\--- `.git` (root git directory)
```

We would like to get the date and time of last change related to the corresponding module directory. 
Аfter the build, we should have N modules with N different timestmps.

TheSnoozer:
Thanks for the detailed explanation and sorry for the delay. I will try to review the changes ASAP.

PR #553, opened 2021-04-21
Reduced code smell

matteobaccan:
Hi

I have cleanup the code for reduce code smell

Action point are derived from sonalLint action items

- Public static field should be constant
- Local variables should not be declared and then immediately returned
- Return of boolean expression should not be wrapped into an if-the-else statement

All the best
matteo baccan


TheSnoozer:
Hi,
cool! Thank you for your contribution. LGTM.

PR #622, opened 2023-05-16
Bring plugins up to date, fix 2 javadoc issues, Require maven 3.2.5

hazendaz:
Brought all plugins up to date.  Adjusted the min maven so it can still run tests to confirm really old end of life maven versions still work.  Not this work is important given maven 3.9.0+ requirements and upcoming maven 4.

The release process will still complain with warnings if using maven 3.9.x.  This has to do with profiles in oss parent that should be removed and directly used as that is long obsolete as stated by nexus.  For now left it with just the in-line build plugins being fully addressed.

It would further be nice to get a release shortly after this although I see possibly some larger change for 6.x, possible that can drop back and be released or go as-is.  Either way this plugin current is vocal due to maven changes which only can be resolved by being up to date and releasing new version.

One note: while GHA works, I was entirely unable to build locally using windows before I even touched any updates.  Tests simply do not work, guessing some other setup must exist first as it complained in regards to the following....

```
Caused by: java.io.FileNotFoundException: File system element for parameter 'source' does not exist: 'src\test\resources\_git_one_commit'
```

All seem similar and there is no file in that location.

### DIFF
--- a/.github/workflows/default-tests.yml
+++ b/.github/workflows/default-tests.yml
@@ -59,7 +59,7 @@ jobs:
     strategy:
       matrix:
         java_version: ['11']
-        maven_version: ['3.1.0-alpha-1', '3.1.1', '3.2.5', '3.3.9', '3.5.4', '3.6.3', '3.8.8', '3.9.1', '4.0.0-alpha-5']
+        maven_version: ['3.2.5', '3.3.9', '3.5.4', '3.6.3', '3.8.8', '3.9.1', '3.9.2', '4.0.0-alpha-5']
 
     steps:
       - uses: actions/checkout@v3
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
 
         <java.target>11</java.target>
 
-        <maven-plugin-api.version>3.9.0</maven-plugin-api.version>
+        <maven-plugin-api.version>3.9.2</maven-plugin-api.version>
         <maven-plugin-plugin.version>3.9.0</maven-plugin-plugin.version>
 
         <junit.version>4.13.2</junit.version>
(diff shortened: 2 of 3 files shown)


hazendaz:
not sure why but my fork was ok, but PR here failed as not compatible with really old end of life versions.  Maven is starting to enforce at least 3.2.5 and I find it hard to believe there is usage before that so just bumping that up.  Also added maven 3.9.2 to build process which is where further issues show up beyond the recent pull requests for similar reasons.

TheSnoozer:
Hello,
thank you for taking the time to update all the plugins!

A ``` 'source' does not exist: 'src\test\resources\_git_one_commit' ```could indicate that you perhaps just have performed a "normal" clone and "forgot" the submodules. This repository has some test-data (actual git-folders) that is used for integrations-tests.

Anyhow, github agrees with your changes and they look very reasonable for me too!

I think what I would like to do, is to release the version as is (so release a 6.0.0) without those changes and only after get them integrated. That way users still have a version if they want to use older maven things, but I agree it's time to drop some old stuff.

Based on https://maven.apache.org/docs/history.html
```
Maven before 3.6.3 has now reached its end of life. New plugin releases will require Maven 3.2.5 or later. The following information is made available for reference.
```

it feels this plugin could then also require a
```
    <prerequisites>
        <maven>[${maven-plugin-api.version},)</maven>
        <maven>[3.8.1,)</maven>
    </prerequisites>
```
but again this is some detail I can take care of. Thanks for getting this started! Will release 6.0.0 now.

TheSnoozer:
Thanks again, 6.0.0 is now released and available. Thus I'm going ahead and merge this for the upcomming 7.0.0 release :-)

Since maven 4.X is still in alpha I'll wait until the first official release.

hazendaz:
Thanks @TheSnoozer .  Main concerns with maven 3.9.x good with 6.0.0 release.   Rest later is good too.  Thank you for quick response here and thanks for pointing out this build has submodules 👍

PR #629, opened 2023-05-22
[pom] Update maven plugins to clear more warnings

hazendaz:
- dependency to 3.6.0
- source to 3.3.0
- checkstyle to 3.3.0
- exec to 3.1.0

This leaves the general mvn clean install build with only javadoc related warning.  Site will still have many unfortunately but its getting better...


TheSnoozer:
Hi, thanks again for this. LGTM.

PR #632, opened 2023-05-31
Update README.md - All configuration options are documented inside `G…

paulwwisl:
…itCommitIdMojo.java` as Javadoc

as said in https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/624#issuecomment-1568765108

> Indeed there is no longer any dedicated "doc" available. I recently moved all the documentation from a dedicated doc to the java code as javadoc.


TheSnoozer:
Sure, sounds reasonable, thanks for adding the additional reference :-)

PR #683, opened 2023-12-27
[pom] Add encoding to pom and correct maven xsd location

hazendaz:
Maven required build xsd settings per maven standards.  This simply adds proper encoding xml marker and corrects site that will get redirected for the xsd.  This has no real impact on the build, just compliance feature.


TheSnoozer:
Thanks for your contribution, LGTM.

PR #700, opened 2024-02-18
Only injects properties into every reactor project once

jeantil:
### Context
Enabling `injectAllReactorProjects` in large and old reactor projects allows to avoid repeatedly parsing the git history to retrieve the information.

In the apache/james-project enabling this property allowed to lower the time spent in git-commit-id from 39s[1] to 2s[2].

However, looking at the verbose logs one can see that the plugin keeps injecting the properties into every reactor project for every project where the plugin runs.

This commit aims to avoid reinjecting keys which have already been set in the context.

[1] https://ge.apache.org/s/xumcl7ztpkmhw/timeline?collapse-all&outcome=success,failed&view=by-type
[2] https://ge.apache.org/s/gpnevfknmdv5a/timeline?collapse-all&hide-timeline&outcome=success,failed&view=by-type

### Contributor Checklist

I'm not saying the fix is perfect, but hopefully it illustrates the issue. If you want a sample project to reproduce you can look at https://github.com/apache/james-project/ enabling verbose output on git-commit-id you will be able to see that for every single project the plugin reinjects all the properties in the 273 projects of the reactor ... effectively setting the properties 273*273 =74529 times.
The performance impact is not huge since its all in memory accesses but it still feels quite wasteful

### DIFF
--- a/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
+++ b/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java
@@ -1234,6 +1234,17 @@ public class GitCommitIdMojo extends AbstractMojo {
         commitIdPropertiesOutputFormat = CommitIdPropertiesOutputFormat.PROPERTIES;
       }
 
+      Properties properties = null;
+      // check if properties have already been injected
+      Properties contextProperties = getContextProperties(project);
+      boolean alreadyInjected = injectAllReactorProjects && contextProperties != null;
+      if (alreadyInjected) {
+        log.info(
+                "injectAllReactorProjects is enabled - attempting to use the already computed values");
+        // makes sure the existing context properties are not mutated
+        properties = new Properties(contextProperties);
+      }
+
       final GitCommitIdPlugin.Callback cb =
           new GitCommitIdPlugin.Callback() {
             @Override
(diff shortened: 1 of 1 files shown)


jeantil:
Wow I never expected this to be reviewed and merge that fast, thanks !

>Perhaps worth to mention that you can also tell the plugin to execute only once by using the [<runOnlyOnce>true</runOnlyOnce>](https://github.com/git-commit-id/git-commit-id-maven-plugin/blob/e7ec7327a9d5e490112ef80fe970f146cb3a15b0/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L710) option.

I did notice about this option, unfortunately it can' t be used for the james project. Most of the projects in the the reactor are made available on apache snapshot repository and released on maven central. To ease diagnostic when a problem occurs we embed a small json file into every jars.
Using `runOnlyOnce` not only resolves the git properties once but also only generates the properties file for the first jar of the reactor. It so happens that the first jar of the reactor is a test utility jar for which the git properties are not as useful as for other jars since they are less likely to have been pulled by james users.
I tried various schemes to generate the file only once and copy it everywhere but maven makes such tasks quite difficult. 

In the end `injectAllReactorProjects` and the properties caching is a good way to solve this :D 

cheers !

LinWanCen:
Can we also merge this feature in version 4. x.x, as many projects use JDK 1.8.

PR #723, opened 2024-03-23
Delete `BigDiffTest`

### LINKED ISSUE #483: Improve BigDiffTest
BigDiffTest takes compared to all other tests a rather long time, for me it needs about 4 minutes.
Apparently most of that time comes from writing 100000 (**!**) files. Could you please check if there are other ways to implement this test?

Long build / testing times make it less attractive for people to run the tests, which could decrease code quality. Additionally writing that many files puts in my opinion an unjustified amount of work on the file system.

Marcono1234:
### Context
Resolves #483

This test was originally added to make sure the `git` process does not hang in case it writes a lot of output to the output streams.

To check this behavior, the test created 100.000 (!) files each a few KBs large and then invoked the plugin, hoping that `git` would produce enough output.
However, this puts a lot of pressure on the file system and disk due to the large number of files. For GitHub CI whose runners are probably using SSDs, this test alone took more than 50% of the time of the complete Maven build. And when running the test with an HDD the test could easily take minutes.

https://github.com/git-commit-id/git-commit-id-plugin-core/pull/101 has added a dedicated and more reliable test which makes sure that the process output is properly consumed, so there is probably no need for this `BigDiffTest` anymore, especially due to the downsides mentioned above.

### Contributor Checklist
- [ ] Ensured that tests pass locally: `mvn clean package`


TheSnoozer:
Mhh, ok why not. Thanks for the contribution!

PR #724, opened 2024-03-23
Use javac `--release` instead of `-source` & `-target`

Marcono1234:
### Context
`--release` also verifies that only API from that Java version is used. Otherwise when building with a JDK newer than the target version it would be possible to accidentally use API not available in the target version.

This also fixes this warning which is currently printed during the build:
> [WARNING] system modules path not set in conjunction with -source 11

### Contributor Checklist


TheSnoozer:
Thanks LGTM!

PR #833, opened 2025-03-09
Update docs

cliserkad:
Adds tutorial code for runtime use of the git properties.
Reorganizes the documentation to present the most important information first.
Some parts were rewritten or removed for conciseness.

Please check out the new versions and click around before passing judgement on the huge diffs.


TheSnoozer:
Thanks for your contribution!
Overall this refactoring looks amazing, thanks for taking the time and reorganizing!

Perhaps two remarks/questions I would need your help in understanding:
1)
Why is there a need for a new workflow `Deploy Javadoc`? Shouldn't the javadoc be automatically build by the process already (via https://github.com/cliserkad/git-commit-id-maven-plugin/blob/cbbacae80b85899de71dc3c9e28e9e2ab7b0152d/pom.xml#L250) also the javadoc is already published to Maven central (e.g. https://repo1.maven.org/maven2/io/github/git-commit-id/git-commit-id-maven-plugin/9.0.1/). Is there perhaps something I'm missing here?

2)
The `docs/using-the-plugin-in-more-depth.md` that was deleted also mentions a custom `Maven resource filtering`, maybe not the end of the world that it's gone...not sure if this something anyone nowadays even uses.


Anyhow, let me emphasize that I think that your changes look already amazing and seem to clean the docs quite significantly, so again huge thanks for taking care of this!

cliserkad:
@TheSnoozer 
The maven workflow commit shouldn't be included in this pull request. I was experimenting with it and didn't expect this pull request to be at the head commit of master.

The generate git properties file option replicates the functionality of maven resource filtering, but requires less configuration. Developers who typically use resource filtering should already know that they can use any maven properties. The readme now clearly states that the plugin exports maven properties, so I think it's a non issue.

I also removed the Spring Bean example because I thought it was out of scope, and the new examples should be very easy to integrate for that use case. Likewise, I think resource filtering is somewhat out of scope.

TheSnoozer:
Thanks for your additional input and your contribution!
I agree, resource filtering is an advanced topic and if I want I could add it to the docs...so let me go ahead and merge.
Thanks again for your work - looks amazing so let me click the merge already!

